Issue #200, opened 2011-12-01
add msg.done()


atmos:
The original hubot had the ability to block execution of other possibly matching `hear` and `respond` calls.  The order of execution was evaluated alphabetically which gets confusing when you have things like `hubot-scripts.json` loading things.  I think we should document the load order behavior and reintroduce this feature.

Script evaluation should look something like this:
1. Process scripts in `scripts/` sorted alphabetically.
2. Process scripts in `hubot-scripts.json` in the order they're defined.
3. Process each match until `msg.done` is true.

This would address a lot of the annoying behavior we see in our campfire when hubot responds to all sorts of random shit ending in a `?`.

PR #218, opened 2011-12-24
add Response.finish() to indicate a message is "handled"

atmos:
- Message now has a @done attribute you can check
- Message and Response both have a finish() method to stop execution
- Addresses #200

Pull Request: master

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name":        "hubot",
-  "version":     "2.0.7",
+  "version":     "2.1.0",
   "author":      "hubot",
   "keywords":    "github hubot campfire bot",
   "description": "A simple helpful Robot for your Company",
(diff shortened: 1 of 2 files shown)


p00j4:
@atmos : i'm having a little trouble here. 
 i'm trying to use this feature but seems like it doesn't work in this case, I need a default response if no match found. but even if match is found, its going in no match found also

```
robot.respond /(start|stop)\s+(.*)/i, (msg) ->
  msg.finish()
  msg.send "match found"

robot.respond /(.*)/i, (msg) ->
   msg.finish()
   msg.reply "sorry, i didn't get you, did you mean any of this? <file_link>"
```

so it prints :
```
sorry, i didn't get you, did you mean any of this? <file_link>"
match found
```

but i'm expecting only "match found" be executed for input "stop this"
though finish seems to work for many cases, but i'm not able to find reason why is not working for this case.

atmos:
@p00j4 Sorry, I stopped writing hubot stuff a while ago.

Issue #556, opened 2013-08-06
External scripts don't work at all


jackmcdade:
I've searched all the docs and blogs on Hubot i could find and dug through the code to figure this out, but i can't seem to make `external-scripts.json` work no matter what i do. 

I have a `*.coffee` file in my local `/scripts` folder, specified it in in `external-scripts.json` file, and upon running Hubot, he dies saying he can't find the npm package.

Is there no way to just write a custom hubot script without putting it on npm or the hubot-scripts repo?

PR #565, opened 2013-08-10
Added script load section to docs

m9:
Adding some docs about how script load works that I was talking about in my [comment](https://github.com/github/hubot/issues/556#issuecomment-22444355)  on #556 issue


drewkhoury:
@balbeko Did this ever make it's way into master? I liked the PR but can't see it in the docs.

technicalpickles:
@drewkhoury see https://github.com/github/hubot/pull/579 which was merged. The documentation has evolved over the past 2 years so it's not in there verbatim anymore, but moved around.

PR #626, opened 2013-12-31
Add x-hub-signature Header Verification to HTTP Requests

joshuarubin:
If an incoming HTTP request has the `x-hub-signature` header set and the `HUB_SECRET` environment variable is set, verify that the signature is valid.

An invalid signature will not be processed as an error will be thrown from the middleware.

If either the `HUB_SECRET` environment variable isn't set or the `x-hub-signature` header wasn't present in the request, then verification will be bypassed. However, `req.hubVerified` will be undefined.

Any script that wants to ensure only verified requests will be processed can do so like this:

``` coffeescript
module.exports = (robot) ->
  robot.router.post "/some/url",
                    robot.requireHubSignature,
                    fn
```

The version of express must also be updated so that it includes the latest connect middleware that provides the verification capability.

### DIFF
--- a/docs/deploying/heroku.md
+++ b/docs/deploying/heroku.md
@@ -49,6 +49,27 @@ configure:
 
     % heroku config:add HEROKU_URL=http://rosemary-britches-123.herokuapp.com
 
+If your hubot will be listening for http connections and you would like to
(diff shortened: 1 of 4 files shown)


bkeepers:
This pull request is quite stale, so I'm going to close it. Better GitHub integration is on the [roadmap](https://github.com/github/hubot/blob/master/ROADMAP.md), so we'll definitely be revisiting this when work on that begins.

PR #650, opened 2014-02-15
Add support for detecting emotes

jasonmay:
I am looking to detect emotes separately from regular messages in my own hubot. I saw on hubot-irc that the author would be willing to support emote detection if an EmoteMessage class was provided in the hubot core, so I decided to take a stab at implementing it. Thanks!


technicalpickles:
This has been open awhile without any forward movement. If anyone is interested  in picking it up, I'd suggest try the new [hubot evolution process](https://github.com/hubotio/evolution/).

PR #655, opened 2014-02-20
How to set up authentication with hubot

jjasghar:
I made a post on my blog, it seems like something that should be in the
docs.


michaelansel:
Closing now that #803 has merged (including guidance for setting up custom authorization policies)

PR #670, opened 2014-03-18
Userforid reconciliation

cozmo:
We were seeing strange failures with our hubot. I was able to trace it down to the `brain.userForId` function, which I then updated to solve the issue. I'm going to outline the failure and the changes made to fix it below. If I'm missing anything as far as why this function was originally coded like it was then I can look at fixing the issue in other ways, but in the context I can see this fix is fairly straightforward. 
### The Issue

Ever since we started using the brain to store info on users we noticed a strange failure where new users weren't having @mention_names assigned. 

What appears to be the problem is initially when users are created (in hipchat) they don't have @mention_names. Hubot adds them to the brain as soon as they exist. Then when @mention_name is set the `userForId` function is called again it doesn't update the brain because of the conditions on [line 109](https://github.com/templaedhel/hubot/compare/github:master...templaedhel:userforid-reconciliation?expand=1#diff-3f212f83b49bd9f3b418259c1f06c0cfL109). 
### The Fix

I removed all the logic around if and when to update users in the brain. Now `userForId` will always return the user for the given ID, as well as updating it with any additional attributes passed in. 

This fixes our issue. 
### Several Questions Remain
- Is there a reason behind the complicated logic around only updating if `options.room` isn't set on the existing user? Tracing back through git didn't dig up anything interesting. I also have trouble seeing how these changes will break any logical uses of this function

### DIFF
--- a/src/brain.coffee
+++ b/src/brain.coffee
@@ -100,17 +100,8 @@ class Brain extends EventEmitter
   # Public: Get a User object given a unique identifier.
   #
   # Returns a User instance of the specified user.
-  userForId: (id, options) ->
-    user = @data.users[id]
-    unless user
-      user = new User id, options
-      @data.users[id] = user
-
-    if options and options.room and (!user.room or user.room isnt options.room)
-      user = new User id, options
-      @data.users[id] = user
-
-    user
+  userForId: (id, options={}) ->
+    @data.users[id] = extend {}, (@data.users[id] or {}), options
 
   # Public: Get a User object given a name.
   #
--- a/src/brain.coffee
+++ b/src/brain.coffee
@@ -100,8 +100,10 @@ class Brain extends EventEmitter
   # Public: Get a User object given a unique identifier.
   #
   # Returns a User instance of the specified user.
-  userForId: (id, options={}) ->
-    @data.users[id] = extend {}, (@data.users[id] or {}), options
+  # If a second argument is passed the used is updated with any attributes contained in that object
+  userForId: (id, options) ->
+    return unless @data.users[id]? or options? #If we're getting a user that doesn't exist short out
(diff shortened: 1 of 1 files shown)


michaelansel:
I did a little digging into the history, and it looks like the conditional was introduced with #282.

> Make sure the undefined user has the correct room set before returning a user object.

Not exactly sure what that means, but it looks like it is related to some since-modified functionality inside `messageRoom`:
- Old: `messageRoom` would try to lookup some kind of user object before sending to a room #208
- New: create a User look-alike instead [robot.coffee](https://github.com/github/hubot/blob/master/src/robot.coffee#L413-L415)

Based on that, I think it is probably safe to completely remove that conditional, but: @elliotttf would you mind dusting off the cobwebs and reminding us of the purpose of your PR?

Once we figure out the desired behavior, we should also add some unit tests that express that behavior.

elliotttf:
Oh wow, that was a long time ago. I *think* it was because users were being added to the brain for rooms that the bot wasn't in but that's a wild guess. There have been a few sleeps since I looked at this :wink:

michaelansel:
@templaedhel would you mind merging master so we can see how this does with all the new automated tests? I think it will only break one test (which we can probably just remove), but I'd like to confirm.

bkeepers:
I tried merging master and running the tests: https://travis-ci.org/github/hubot/jobs/234339545

```
  1) Brain Unit Tests #userForId does an exact match:
     AssertionError: expected undefined to not equal undefined
    at Context.<anonymous> (test/brain_test.coffee:149:9)
  
  2) Brain Unit Tests #userForId when there is no matching user ID creates a new User:
     AssertionError: expected undefined to be an instance of User
    at Context.<anonymous> (test/brain_test.coffee:168:11)
```

PR #676, opened 2014-04-01
Clean up events during shutdown

aaronstaves:
#### Summary
Looks like instantiating a new Robot will attach an 'uncaughtException' handler to the main process object using an anonymous function.  This listener is never cleaned up, even when shutdown is explicitly called.  I've added a function emitError that can be referenced when removing the event listener.

#### Issue
The issue is that when a robot is created and shut down multiple times (my use case is unit tests, but there may be others) the event listener for 'uncaughtException' is never removed.  I'll specifically receive a warning in this when running my unit tests, you can see that in the travis-ci link below.  Feel free to use my plugin to reproduce.

#### Links 
Unit Tests with warning - https://travis-ci.org/aaronstaves/hubot-substitute/jobs/21996378.
Hubot script - https://github.com/aaronstaves/hubot-substitute
NPM - https://www.npmjs.org/package/hubot-substitute

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -63,8 +63,7 @@ class Robot
 
     @on 'error', (err, msg) =>
       @invokeErrorHandlers(err, msg)
-    process.on 'uncaughtException', (err) =>
-      @emit 'error', err
+    process.on 'uncaughtException', @emitError
 
 
   # Public: Adds a Listener that attempts to match incoming messages based on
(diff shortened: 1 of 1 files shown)


technicalpickles:
I merged https://github.com/github/hubot/pull/848 because it was merging cleanly. Thanks for contributing, and sorry it took so long to address!

Issue #673, opened 2014-03-24
Question:  Preventing Concurrent Access?


ajkerr:
Hi there,

Is it possible to configure Hubot to prevent concurrent commands?  We have certain commands that can we run in Campfire that can clobber each other if run simultaneously.

I don't see anything obvious in the source code, but would appreciate any guidelines on how best to achieve this.

Thanks,
Andrew

PR #678, opened 2014-04-02
Add documentation on how you can prevent concurrent execution of your scripts.

ajkerr:
Added documentation based on results of discussion in #673.


mose:
@technicalpickles is this clear for merge ? There is a minor conflict I can solve it easily.

technicalpickles:
If anyone is feeling ambitious, this would make a good listener middleware, which ensures single concurrency of listeners with a particular piece of metadata.

PR #684, opened 2014-04-16
Added Redis server to the install documentation for windows users.

etoxin:


bkeepers:
@etoxin thanks for the pull request. I think I'd rather we improve the on-boarding experience with Hubot and not require redis out of the box. I made a not of it and will make sure it gets addressed soon.

PR #685, opened 2014-04-16
Reduce duplicated regex patterns, implement @escapeRegExp private method...

hbakhtiyor:
... for escaping regex.

Use yaml style library declaration.


technicalpickles:
It's been awhile, and this doesn't seem to cleanly merge anymore :sweat: If you are still interested, can you split up the duplicate regex from the style changes in a new PR?

PR #686, opened 2014-04-17
Add an 'ignore' method to the robot

dancrumb:
I'd like my bot to be able to stop listening and hearing things.

Thus, I'd like an 'ignore' method, or an ability to get something back from `hear` and `listen` that I can then call `stop` on, to stop hearing and responding.


michaelansel:
More cross-linking; listener de-registration could be a useful construct to include in a new script interface (#858).

I like the idea of `robot.hear` (and friends) returning a usable handle, though I wonder if we can make that the Listener object instead of a proxy.

Also, I wonder if this could be implemented with middleware: retrieve a listener by ID and then set metadata to "disabled", which a piece of listener middleware would then use as an indicator to fail.

Rough idea for middleware:

The middleware:
```coffeescript
module.exports = (robot) ->
  robot.listenerMiddleware (robot, listener, response, next, done) ->
    if listener.options.ignore?
      # Block listener execution
      done()
    else
      # Continue
      next(done)
```

A script that toggles itself on and off:
```coffeescript
module.exports = (robot) ->
  robot.hear /message123/, id: 'my-cool-listener', (response) ->
    # time to turn off
    robot.listenerById('my-cool-listener').options.ignore = true
    # ....
    # time to turn back on
    robot.listenerById('my-cool-listener').options.ignore = false
```

Depends on #803 and `robot.listenerById` (https://github.com/github/hubot/issues/823#issuecomment-73999270)

michaelansel:
Opened #1031 to add `robot.listenerById`. When that merges, I'll close this.

bkeepers:
Since it looks like #1031 is the preferred way of handling this, I'm going to go ahead and close this. Thanks for your contributions!

PR #698, opened 2014-05-14
Added recent messages function in campfire adapter.

milyh:
This function returns recent message from current room.
Params:
  limit: number of messages returned
  callback: result
Example:
  robot.adapter.bot.Room(msg.message.room).recentMessages 2, (error, data) ->
    msg.send data

Campfire API (recent messages): https://github.com/basecamp/campfire-api/blob/master/sections/messages.md#get-recent-messages


bkeepers:
Hey @milyh, I agree with @technicalpickles' [comment](https://github.com/github/hubot/pull/698#issuecomment-43520013) that I don't think we should add methods to adapters that aren't used internally, so I'm going to close this. Thanks for taking time to contribute though!

PR #700, opened 2014-05-15
Campfire Adapter emit errors

technicalpickles:
We've had the ability to `emit 'errors'` for awhile, but the Campfire adapter wasn't updated to use this. This should mean that if you have an error handling script, you'll be able to catch the floods of rate-limiting errors when you connect with lots of rooms.

cc https://github.com/github/hubot/pull/699

### DIFF
--- a/src/adapters/campfire.coffee
+++ b/src/adapters/campfire.coffee
@@ -238,6 +238,7 @@ class CampfireStreaming extends EventEmitter
                     data.body
                   )
                 catch error
+                  @robot.emit 'error', error
                   logger.error "Campfire data error: #{error}\n#{error.stack}"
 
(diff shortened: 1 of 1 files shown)


technicalpickles:
It is probably okay to merge, but it has been so long I don't remember what the underlying problem we were trying to identify, and now that we're off Campfire, there's not a good way to validate it.

bkeepers:
Ok. Should we just close this then?

mose:
Code seems to be safe enough, just covering usage of `err`. I would vote for a merge.

technicalpickles:
I'm just going to close this. I don't think it's a good habit of getting into of merging code that "looks good" without actually being able to validate it.

PR #715, opened 2014-06-13
changes to robot.coffee to invalidate require cache on loadfile.

srobroek:
This change adds a require.cache invalidate to robot.coffee's loadfile function. With this addition, reloading scripts becomes possible (through reload.coffee or script.coffee) without restarting the hubot process.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -211,6 +211,13 @@ class Robot
   loadFile: (path, file) ->
     ext  = Path.extname file
     full = Path.join path, Path.basename(file, ext)
+    if  require.cache[require.resolve(full)]
+      try
+        cacheobj = require.resolve(full)
+        @logger.debug "require cache for #{cacheobj} invalidated."
+        delete require.cache[cacheobj]
+      catch error
+        @logger.error "Unable to invalidate #{cacheobj}: #{error.stack}"
     if require.extensions[ext]
       try
         require(full) @


ileathan:
Merge please.

michaelansel:
Script reloading overall feels pretty sketchy.

Looking at [reload.coffee](https://github.com/github/hubot-scripts/blob/master/src/scripts/reload.coffee), it currently operates by digging into the Robot object and trying to rip out all state created by loading a script (it gets the standard stuff, but misses all the other things a script could set up (it is common to attach new things to the Robot object and #803 will add an entirely new variable)). This is pretty fragile and (I personally think) shouldn't be supported with changes to the core (that are, in turn, dependent on the internals of node.js).

In addition, flushing the require cache is going to make unit testing scripts trickier than it already is (right now, you can preload a script and plug in some mocks before running robot.loadFile).

If there is a strong need for script reloading without restarting, I think proper namespacing/sandboxing of scripts is the best option so that we can easily remove everything touched by a script. This may be something we can fit in as part of #858.

@technicalpickles I think this one is up to you to decide if hot-reloading of scripts should be supported.

technicalpickles:
It is definitely a bit tricky, because scripts can make changes to the robot object if they want. That is to say, it'd be hard to make sure all scripts can be reloaded gracefully.

I think the thing to think about is what we are trying to get script reloading. I can appreciate wanting to do it in, say, development, in order to have a faster feedback cycle. There may be other ways of achieving that.

michaelansel:
Could we instead focus on accelerating the startup process? Then, you can just `while true ; do bin/hubot ; done` and `hubot die` whenever you want a reload.

Adding automated test guidance could also address a lot of this (working under the assumption that anything you would test by hand could be written as an automated test).

ileathan:
> That is to say, it'd be hard to make sure all scripts can be reloaded gracefully.

That should be up to the author of the script otherwise its a cat n mouse game.

I sure hope this gets supported since i have scripts that automatically rebuild the bots code all the time.

Not sure how people can stand 
> the startup process


I still dont see why you wouldnt just merge this but ill just create a fork and point to it instead.

michaelansel:
@ileathan Can you explain more about your use case? As @technicalpickles mentioned, supporting hot-reloading of scripts in a predictable way will be tricky. As a user, I shouldn't need to understand a script's implementation to know whether or not I can reload it without corrupting the bot.

srobroek:
@michaelansel i'll explain you my use case for hot reloading scripts. We use hubot as an irc bot, and since we have multiple people maintaining their own scripts reloading those without a hot reload means having the bot disconnect and reconnect on a regular basis. In addition, currently these people can just give hubot a "reload all scripts" command if their script works in dev/test instead of having to either schedule a full reload or give these developers some way to restart the whole process. 

Since we only use very simple scripts (mainly HTTP API calls) i don't believe there's any way to cause any data corruption. Maybe we could have a startup flag with a warning that either allows or disallows hot reloading?

ileathan:
Same this but more specific one script

ileathan:
Either way @srobroek's fix did the trick +10

srobroek:
well, as the original submitter of the pull request i thought it'd be appropriate to finally comment on this :)

rushi:
I found it quite useful while developing new scripts for hubot. I used the shell adapter, wrote my code and just reloaded the scripts to keep testing new functionality

michaelansel:
Thanks all for the feedback. I'm still staring at this and trying to work out a good solution. I *think* the right answer here is to treat your robot more like a software project (automated instead of manual testing, packaged releases instead of hot-reloading random components), but I recognize that could be rather disruptive and therefore want to think more about the implications before making that suggestion.

Immediate thoughts:
- we need a clear, easy-to-use test pattern for scripts (right now it is fairly clunky); I'd like to try to get this before #858, but I'm not sure if we can resolve all the funkiness without changing the script interface
- restarts are annoying for heavily used bots; it would be awesome if we could dump all state and reload the robot while queuing up messages in the adapter; this may not happen until we have #858

ileathan:
Personally i still really like srobroek's solution. my bot is trying to recode itself and i usually dont care for the automated tests WHATSOEVER.

I also didnt like #858 and hope it doesnt get implemented :(

But thanks so much for the hard work :+1:

src/robot.coffee
```diff
@@ -211,6 +211,13 @@ class Robot
   loadFile: (path, file) ->
     ext  = Path.extname file
     full = Path.join path, Path.basename(file, ext)
+    if  require.cache[require.resolve(full)]
```
technicalpickles: Should we put this conditional under `if require.extensions[ext]` below? The require isn't attempted until then, so if there's not going to be anything to invalidate if it's an extention that's not supported.

src/robot.coffee
```diff
@@ -211,6 +211,13 @@ class Robot
...
+      try
+        cacheobj = require.resolve(full)
+        @logger.debug "require cache for #{cacheobj} invalidated."
+        delete require.cache[cacheobj]
+      catch error
+        @logger.error "Unable to invalidate #{cacheobj}: #{error.stack}"
```
technicalpickles: Have you ever run into this error? I'm wondering if it's okay to just log it, or if it should do anything else if it failed to invalidate.

ileathan: Not once no

srobroek: I personally haven't, but i considered that it would be possible for whatever reason which is the reason for the try/catch. @technicalpickles  i'd be fine with moving this below the require.extensions[ext], but since there's no require cache generated at initial load the if won't ever be true during the initial load. 

So i think this'd mainly be a cosmetic change instead of a functional one.

technicalpickles: Could we move this into the `try` block below instead? That way, it's only one try/catch block, it's logically grouped under the `if require.extensions[ext]`, and it also will use the hubot's error handling to report any issues with clearing the cache (ie `emit 'error'`).

rothgar:
@srobroek please update so this can be merged. Would be very handy. I manually patched my instances and it's working great

ileathan:
I noticed that in the future people do not pay the github workers they pay the kids to write tutorials. Which is fine however I believe if we allow marking of issues with bits then eventually the kids writing the tutorials will become freinds with the source code maintainers. Once they are freinds I think they will naturally work together for more econ++

Otherwise srobroek i recommend you not work for free... because they hire kids to get paid for copy pasting your work (which is fine if they also pay the author imo)

ileathan:
Sorry for being political im just homeless in 23 hrs and very depressed, i wont post again on here unless its technical or more related.

sorry

PR #726, opened 2014-07-01
Make use of HUBOT_PATH env variable when loading scripts

enykeev:
Currently, you have to be in a proper working directory to launch Hubot. At the same time, we have defined HUBOT_PATH we are almost not using at all.

After that patch, you would be able to execute Hubot from any directory by providing proper HUBOT_PATH:

``` sh
BASEDIR=path/to/bot

npm install --prefix $BASEDIR

env HUBOT_PATH="$BASEDIR" $BASEDIR/node_modules/.bin/hubot
```

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -81,26 +81,26 @@ else
   robot.alias = Options.alias
 
   loadScripts = ->
-    scriptsPath = Path.resolve ".", "scripts"
+    scriptsPath = Path.resolve Options.path, "scripts"
     robot.load scriptsPath
 
-    scriptsPath = Path.resolve ".", "src", "scripts"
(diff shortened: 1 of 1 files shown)


technicalpickles:
As part of this, we'd also want to make the shell adapter look use the HUBOT_PATH in order to know where to put the .hubot_history.

enykeev:
I don't think I will be. Haven't touched hubot code since the commit and have to reason to get back to it any time soon.

bkeepers:
Ok, thanks @enykeev. I'm going to go ahead and close this for now since we're mostly focused on cleaning up old issues and pull requests and we can revisit this if it comes up again later.

PR #744, opened 2014-08-15
add functionality of logging to file

tristanguo:
Hi guys,

I am feeling it is quite useful to make logs be persisted to disk files rather that only print to stdout. Especially when I also want to use robot.logger to log some script layer information, I would like to have it be kept on file and with hubot internal logging information.

So the easiest way of achieving this is simply use a global variable HUBOT_LOG_FILE, if it is specified, then we will construct a Log instance with a file stream.

Hope this make sense to you, Thanks a lot!

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -10,15 +10,17 @@ OptParse = require 'optparse'
 Path     = require 'path'
 
 Switches = [
-  [ "-a", "--adapter ADAPTER", "The Adapter to use" ],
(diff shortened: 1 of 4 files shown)


atonparker:
Hey what's the status of this? It'd be great to see feature this released.

technicalpickles:
Nothing new :sweat:  Last status was I asked for updated documentation about how to log from scripts.

bkeepers:
Updated PR in #1164

PR #752, opened 2014-08-29
fix #619: brain loading timing issue

### LINKED ISSUE #619: Brain loading timing issue
Hi,

I was trying to add some user's information to the brain when hubot-xmpp receives presence messages. However, my setup uses redis-brain and there is a timming issue with the redis initial brain load.

The problem is that hubot-xmpp receives presence informations and stores that information into the brain before redis-brain load the initial brain content. When redis finally load the data, it "merges" the data into the brain. That actually override everything in the "users" key without further notice.

I looked at both redis-brain and hubot code and I don't see any easy way to solve that. It seems to me that the brain should be loaded before the adapters are started but I would like to fetch some advices before I attempt to patch/submit a pull request.

One possible solution would be to force brain implementation to extend the Brain class and add a switch to the hubot startup to specify the brain to use. This would be similar to the adapter. However, this seems pretty invasive and imply that you can't have two concurrent brain backend loaded at the same time (which is not necessarily a bad thing when you consider that loading data from two brain is not really supported anyway)

Any idea?

conradkdotcom:
Following the discussion in #619, here's a fix. I've tested it with the `hubot-plusplus` module, which was my use case. It works in that case.

If there are any questions about it, please let me know :-)
### Possible BC breaks

If devs rely on there being a timing issue in the Hubot core, some scripts might break as they might do something like:

``` coffee
if robot.brain.data
    doStuff(robot.brain.data)
robot.brain.on 'loaded', (data) ->
    doStuff(data)
```

In the case described above, with this fix, the `on loaded` event listener would effectively be called twice, whereas before, it would have been called only once.

/cc @technicalpickles

### DIFF
--- a/src/brain.coffee
+++ b/src/brain.coffee
@@ -148,6 +148,30 @@ class Brain extends EventEmitter
 
     matchedUsers
 
+  # Public: Override the base EventEmitter.on method to handle the 'loaded'
+  # event even if it was emitted before registering the event listener. This is
+  # needed to avoid timing issues. See:
+  # https://github.com/github/hubot/issues/619
+  #
+  # Returns nothing.
+  on: (event, listener) ->
+    if event is 'loaded' and @loadedEventEmitted
+      listener(@data)
+    else
+      EventEmitter.prototype.on.call this, event, listener
+
+  # Public: Override the base EventEmitter.emit method to handle the 'loaded'
+  # event even if it was emitted before registering the event listener. This is
+  # needed to avoid timing issues. See:
+  # https://github.com/github/hubot/issues/619
+  #
+  # Returns nothing.
+  emit: (args...) ->
+    event = args[0]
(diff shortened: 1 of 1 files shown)


michaelansel:
@conradkleinespel We now have a pretty good testing framework in place. Any chance you would be willing to put in some automated tests for this? For timing-dependent test examples, take a look at the Brain tests: https://github.com/github/hubot/pull/903/files#diff-8fa2d81c11c765b2f0fbcd0d54ccf3f8

conradkleinespel:
@michaelansel Back then, I would have loved too, as Hubot is an interesting project. Unfortunately, I don't have the time right now, as I have a lot of other stuff going on. Best of luck though !

michaelansel:
Did a little noodling and I think adding a new function like `robot.brain.onLoaded` might be better than intercepting calls to EventEmitter. I'll keep noodling...

PR #772, opened 2014-09-23
Remove express

therealklanni:
Removing express from hubot in preparation for [hubot-express](http://github.com/hubot-scripts/hubot-express) package.


mose:
Should we close this and consider having a discussion about express inclusion/exclusion in hubot 3? Seems the better course of action ...cc @technicalpickles @bkeepers

bkeepers:
> Should we close this and consider having a discussion about express inclusion/exclusion in hubot 3? Seems the better course of action ...cc @technicalpickles @bkeepers

👍 I'm good with that. Looks like there are quite a few merge conflicts here, so let's take some time to discuss it before asking anyone to do work to fix them.

Issue #767, opened 2014-09-18
Extract express out of hubot core


technicalpickles:
The main problem with hubot's express support, from a user's perspective, is that it doesn't allow a lot of customizability. Sure, you can add routes to it easily enough, but if you need to add or swap middleware, it's less obvious how to do that in a clean way.

A problem from a hubot maintenance point of view is that it's a part of hubot's core that people want to change, so any changes to it have to be carefully considered before merging or it could impact existing users.

So the general thinking has been to remove `robot.router` from hubot-core, and move it to its own package. The main obstacle to doing that is... if express support is external to core, how would scripts go about adding their own routes, middleware, etc, without clobbering others? In addition, there's not a clear path for something like this being an extension to hubot, rather than a package of scripts.

Most recent discussion over in https://github.com/github/hubot/pull/764

mose:
Should we consider to leave express in hubot 2 and just postpone this thinking for hubot 3?

technicalpickles:
I have been thinking about this a bit, and I don't think we can quite remove express, as it will be useful when building adapters that use webhooks.

The main motivation for me to want to remove express was that there have been many requests over the year to adjust how the builtin express app works, and that usually involves exposing environment variables to turn things on or off. There are an infinite number of combinations though, so it doesn't really scale.


Instead of removing, maybe we could do something similar to botkit, which has a [default way to setup express, or you can pass in your own instance of express](https://github.com/howdyai/botkit/tree/master/docs#use-botkit-with-an-express-web-server). I think that would give users a lot more flexibility, while offering express as something out of the box.

I think the main blocker to doing that is https://github.com/github/hubot/issues/858

technicalpickles:
Assigning myself to write up a [Hubot Evolution](https://github.com/hubotio/evolution/) proposal for this.

timkinnane:
@technicalpickles That suggestion of passing in custom instance of Express sounds great, can see it being super useful for who knows what edge cases are out there, while keeping the core support intact, which I think is also very useful for the broader majority.

I will also use this opportunity to politely request again that stale bot be executed. Or at least stuffed in a basement until core team activity resumes.

therealklanni:
@timkinnane @technicalpickles I think that was at least in part the driver behind refactoring Express out, so I think this is a reasonable solution.

PR #775, opened 2014-09-24
Add deprecation warning for inbuilt Express

therealklanni:
The title says it all.

@technicalpickles is there more we should say/do to deprecate this in the current minor version?

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -51,6 +51,7 @@ class Robot
 
     @parseVersion()
     if httpd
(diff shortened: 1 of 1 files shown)


mose:
I don't think this PR is relevant anymore if we postpone the decision about express to hubot 3.

technicalpickles:
We closed https://github.com/github/hubot/pull/772 so I think we can close this one. I added `triage: evolution` to https://github.com/github/hubot/issues/767 where all of this started.

Issue #768, opened 2014-09-18
Provide method for external authorization policy


michaelansel:
Building off of the filtering mechanism in #724, I'm trying to build a mechanism for setting an authorization policy outside of individual scripts. I want to do this so I can specify my own policy around who can execute commands (and potentially what level of identify verification is required) without forking every script and injecting my own company-specific code.

I think the rough end-result is something like RBAC with a permissions file that maps individual listeners to permitted users/groups. In that model, I need a way to reliably reference individual listeners so that I can apply policy to them.

The most precise, but least stable, method I can think of would be to reference each listener by regex. This feels very dirty though...

The most stable method would be to add an explicit, unique identifier to every command. This could be accomplished by tweaking the respond/hear API to include an identifier field and putting together a PR for every script I intend to use. Not a terrible solution, but it would be a lot of work and requires lots of upstream acceptance.

An in-between option would be to specify an example message and then search for all listeners that match the message. This isn't perfect, but accomplish the goal well enough.

I'm sure other concerns will come up along the way, so opening the issue now to start collecting feedback and posting updates as I work on this.

technicalpickles:
> Going to leave this open until the listener middleware merges.

It's been merged for awhile 😂

PR #801, opened 2014-10-23
Named Listeners

michaelansel:
As the first part of enabling #768, we need a way to label each Listener with a scoped name. By itself, this first PR is fairly useless (no way to access the ID), but I want to work on this in bite-sized chunks.

I've updated the documentation to show that you can (and should) specify an ID for all Listeners. Once this gets merged and the RBAC module is created, I'll do a pass over hubot-scripts and file a bunch of PRs for adding Listener IDs and default Permissions.


michaelansel:
Reopening this PR to make some forward progress and make it clear what is happening in #803

bhuga:
:sunglasses: :boat:

PR #803, opened 2014-10-25
Listener Middleware

michaelansel:
Builds on #801 (57b722022807b9a5befaafcab2442ad3a03d455b) to allow injecting arbitrary code between the Listener match and execute steps. Like in Express, middleware can interrupt the response process, preventing the Listener callback from ever being executed. Middleware can perform operations both on the way towards the Listener callback (before callback execution) and on the way away from the Listener callback (after callback execution/middleware interrupt).

As a side effect, listeners are now executed asynchronously. Behavior around message.done should remain the same (process until message.done is true).

Example usage (imposes authorization policy on all scripts): https://github.com/michaelansel/hubot-rbac/blob/master/src/rbac.coffee

API Change:
* `Robot.receive` and `Listener.call` are now asynchronous. Any external code depending on knowing when all Listeners have been tested (in `receive`) or when the Listener callback has finished executing (in `call`) will need to be updated to use the new callback parameter. Code that doesn't care about timing should continue to work without modification.


michaelansel:
@technicalpickles, is there anything I can do to improve your confidence in this PR? I'm happy to improve documentation, refactor the code, etc. if you can point me to areas you are uncomfortable with.

michaelansel:
Automated tests are on the way! As soon as #841 merges, I'll write tests for all the code modified here in a separate PR (tests that pass against the pre-middleware code). Once those tests are merged in, I'll merge master back into this PR and adjust as necessary to make everything pass once again (I expect a few small changes will be necessary in the unit tests).

michaelansel:
Tests have been updated to match the changes in middleware (basically make everything async); I'll do all the necessary merging/rebasing once #841 has merged. Next, I'll start working on adding new tests to cover the net-new middleware code.

https://github.com/michaelansel/hubot/compare/middleware-with-testing...middleware-tests

technicalpickles:
@michaelansel I haven't forgotten about this, just a little bit slow :sweat:

michaelansel:
This looks like the same failure fixed by #869, but the fix doesn't appear to be working... @afeld any ideas?

afeld:
Hmm, [other builds](https://travis-ci.org/github/hubot/builds) have passed... maybe [the `npm update -g npm`](https://travis-ci.org/github/hubot/jobs/51110617#L95) failed? No error message, which is weird. Can someone w/ write access to the repo try [restarting the build](http://stackoverflow.com/a/21727193/358804)?

If [updating NPM in the build](https://github.com/github/hubot/pull/869) proves to be too flaky, we change the package.json to be NPM 1.2-compatible.

technicalpickles:
Restarted it: https://travis-ci.org/github/hubot/builds/51110614

afeld:
See? Fixed!

back away slowly

michaelansel:
Sadness @ flaky tests :-(

michaelansel:
As part of writing the tests, I realized two things:
1. Listener is tightly coupled to the implementation of Robot (directly accesses `@robot.middleware.listener`)
2. `executeAllMiddleware` is pretty generic

So, I'm thinking about moving `executeAllMiddleware` over to Robot and calling it with the name of the set of middleware that should be executed (in this case, "listener"). I also want to make sure there is sufficient test coverage to ensure I'm not breaking anything.

michaelansel:
Added some functional tests (so that I wasn't modifying all the tests and all the code at the same time) and then moved the middleware execution logic over to Robot. This also required some changes to internal APIs. /me loves automated testing sooo much right now.

Due to extensive middleware/listener error handling, I also added some rethrow logic to give clear failure messages instead of the errors just getting swallowed.

docs/scripting.md
```diff
@@ -598,3 +602,81 @@ module.exports = (robot) ->
...
+    # code to stop annoying someone
+```
+
+These scoped identifiers allow you to externally specify new behaviors like:
+- authorization policy: "allow everyone in the `annoyers` group to execute `annoyance.*` commands"
+- rate limiting: "only allow executing `annoyance.start` once every 30 minutes"
+
+# Middleware
+
+Hubot supports injecting arbitrary code in between the listener match and execute steps. This allows you to create very interesting extensions that apply to all scripts. Examples include centralized authorization policies, rate limiting, logging, and metrics. Middleware is implemented just like any other hubot script, however instead of using the `hear` and `respond` methods, middleware is registered using `listenerMiddleware`.
```
michaelansel: changing to "inserting additional logic"

michaelansel:
Did some reading through old comments (https://github.com/github/hubot/pull/803#issuecomment-64054374) and decided to add some more test cases that explicitly cover the middleware public API, specifically the exposed data structures. This should ensure we don't accidentally break all middleware by making internal changes. The tests are pretty simple because the most complicated objects (robot, response) are already exposed to listeners, and I didn't want to conflate testing their APIs with testing the middleware API. Explicit Robot and Response tests will probably evolve over time as I have time (and understand their interfaces better).

michaelansel:
Found a regression in behavior related to listener callback exceptions :disappointed:

Added a functional test to prove the regression (passes on pre-middleware code, was breaking on middleware), and then added some unit tests and the fix.

michaelansel:
I stumbled across the node docs for `process.nextTick` and was intrigued by the comments on sync vs async APIs. Specifically, a function with a callback should either be 100% synchronous (callback executed prior to function returning) or 100% asynchronous (callback executed no sooner than next turn of the event loop). I'm not sure where middleware falls, but I'll do a run through, make corrections, and add test cases that ensure we are consistent.

src/listener.coffee
```diff
@@ -1,33 +1,79 @@
...
   #
   # message - A Message instance.
   #
-  # Returns a boolean of whether the matcher matched.
-  call: (message) ->
+  # @callback - Call with a boolean of whether the matcher matched.
+  #
+  # Returns nothing
+  # Returns before executing callback
+  call: (message, cb) ->
```
technicalpickles: So, this is a public function that is changing semantics. I think that means either:

* change to private, since it's not really used outside of hubot
* update to support running without a callback
* do a major version bump when releasing

michaelansel: Whoops. I'll update to not break the old behavior. The only breakage that *could* happen is if a caller was expecting the listener callback to begin execution before this function returns. Given that this is both an edge case and timing was never clearly specified, I think we are okay there.

I think this is probably marked as public because it is public from the class's perspective (called by Robot). We should probably look at having Hubot public APIs vs shared internal APIs, or something like that...

technicalpickles:
Been thinking how I can feel more confident in releasing this. So, started making some changes to feel more comfortable with the code, and adjusted some of the docs.

I also released a beta based on this branch as '2.12.0-beta' if anyone wants to test it.

After releasing that, I started porting a script we had used to black list specific users. It basically looked like:

```coffeescript
  BLACKLISTED = [
    "000000000" # Enter ids here, as strings
  ]

  robot.respond /.*/i, (msg) ->
    if msg.message.user.id.toString() in BLACKLISTED
      msg.reply "I'm sorry @#{msg.message.user.name}, but I'm configured to ignore your commands."
      msg.finish()
```

As a middleware:

```coffeescript
  BLACKLISTED = [
    "000000000" # Enter campfire ids here, as strings
  ]

  robot.listenerMiddleware (robot, listener, response, next, done) ->
    if response.message.user.id.toString() in BLACKLISTED
      # console.log listener.regex
      response.reply "I'm sorry @#{response.message.user.name}, but I'm configured to ignore your commands."
      done()
    else
      next(done)
```

This works correctly to block respond commands, but I the `response.reply` ended up happening multiple times. That was because we had a few `robot.hear /.+/`, and the listener middleware was firing for both those and responds.

I'm not sure that's a blocker yet, but I think this might hint at complications when adding listeners to an existing hubot project.

michaelansel:
@technicalpickles I'm working on [documentation](https://github.com/michaelansel/hubot/compare/middleware...michaelansel:middleware-idea?name=middleware-idea) right now to explain authorization patterns. I think it might make more sense to make the linked doc changes after middleware merges (just so we don't keep piling things onto this single PR), but it might be useful for you in your testing right now.

Another note: yes, building middleware is tricky because it is executed for *every* listener that matches a message. Definitely more complicated than building a normal listener.

technicalpickles:
> I also released a beta based on this branch as '2.12.0-beta' if anyone wants to test it.

I had to unpublish this, because it was being used by brand new hubots. Going to have to figure out the beta way of publishing things now, because the way I did it obviously was incorrect.

michaelansel:
Realized I should probably throw this in a comment in addition to updating the description:

> API Change:
>
> `Robot.receive` and `Listener.call` are now asynchronous. Any external code depending on knowing when all Listeners have been tested (in `receive`) or when the Listener callback has finished executing (in `call`) will need to be updated to use the new callback parameter. Code that doesn't care about timing should continue to work without modification.

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+  #
+  # Returns nothing.
+  listenerMiddleware: (middleware) ->
+    if middleware.length != 5
+      throw new Error("Incorrect number of arguments for middleware callback (expected 5, got #{middleware.length})")
+    @middleware.listener.push (robot, context, next, done) ->
+      middleware.call(undefined, robot, context.listener, context.response, next, done)
+    return undefined
+
+  # Execute all middleware in order and call 'next' with the latest 'done'
```
technicalpickles: Let's mark this as `Private`. I don't think it's something a script writer would need to be using at all.

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+  # callback if last middleware calls through. If all middleware is compliant,
+  # 'done' should be called with no arguments when the entire round trip is
+  # complete.
+  #
+  # whichMiddleware - String name of middleware set to execute (currently
```
technicalpickles: Very nice! I totally missed that you made this a more generic.

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+  #
+  # next(context, done) - Called when all middleware is complete (assuming
+  #     all continued by calling respective 'next' functions)
+  #
+  # done() - Initial (final) completion callback. May be wrapped by
+  #     executed middleware.
+  #
+  # Returns nothing
+  # Returns before executing any middleware
+  executeMiddleware: (whichMiddleware, context, next, done) ->
```
technicalpickles: maybe middlewareType might be a better name than whichMiddleware?

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+        # Forcibly fail the middleware and stop executing deeper
+        doneFunc()
+
+    # Executed when the middleware stack is finished
+    allDone = (_, finalDoneFunc) -> next(context, finalDoneFunc)
+
+    # Execute each piece of middleware, collecting the latest 'done' callback
+    # at each step.
+    process.nextTick ->
+      async.reduce(allMiddleware, done, executeMiddleware, allDone)
```
technicalpickles: `reduce` is the right thing, but I basically have to refer back to the documentation every time. I'm wondering if we can either add a code comment to explain done = memo, executeMiddleware = iterator, allDone = callback, or maybe we can just assign a local variables for those for readability?

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+  # done() - Initial (final) completion callback. May be wrapped by
+  #     executed middleware.
+  #
+  # Returns nothing
+  # Returns before executing any middleware
+  executeMiddleware: (whichMiddleware, context, next, done) ->
+    if not @middleware.hasOwnProperty whichMiddleware
+      throw new Error "Invalid middleware set: \"#{whichMiddleware}\""
+
+    allMiddleware = @middleware[whichMiddleware]
```
technicalpickles: Maybe just `middleware`? That wouldn't interfere with `@middleware`

src/robot.coffee
```diff
@@ -184,24 +190,112 @@ class Robot
...
+
+    # Execute a single piece of middleware and update the completion callback
+    # (each piece of middleware can wrap the 'done' callback with additional
+    # logic).
+    executeMiddleware = (doneFunc, middlewareFunc, cb) =>
```
technicalpickles: Having this variable `executeMiddleware` inside a function `executeMiddleware` is a little confusing.

technicalpickles:
I was doing some more research into middleware in general, and came across [The strange world of Node.js design patterns](http://www.mariocasciaro.me/the-strange-world-of-node-js-design-patterns), which mentions the middleware pattern. That gave me some inspiration to dive deeper into the inspiration.

I left a bunch of smaller comments, but I had two bigger ideas:

I saw some methods take a `context` object which has the `listener` and `response`. What if the `robot.listenerMiddleware` API took a context object instead of a `listener` and a `response` separately? I think that would help future proof the API a bit if we ever wanted to add anything else that's passed along.

I was also seeing a lot of interactions back and forth between robot to listener to robot. Do you think making a middleware class, or possibly a middleware stack class, would help with these interactions? That might simplify things, where a robot has a middleware stack, and listeners, and each listener just knows it has to go through the middleware stack.

technicalpickles:
> I stumbled across the node docs for process.nextTick and was intrigued by the comments on sync vs async APIs. Specifically, a function with a callback should either be 100% synchronous (callback executed prior to function returning) or 100% asynchronous (callback executed no sooner than next turn of the event loop). I'm not sure where middleware falls, but I'll do a run through, make corrections, and add test cases that ensure we are consistent.

I was trying to read up more on this too, and found this article useful: http://howtonode.org/understanding-process-next-tick

src/robot.coffee
michaelansel: How are we defining public vs private? Is it for the module or the class? Once we have a decision, I think some of the classifications need to change for different methods.

src/robot.coffee
michaelansel: Sure, will do.

src/robot.coffee
michaelansel: Can do.

src/robot.coffee
michaelansel: Make sense. Missed that during the refactor.

src/robot.coffee
michaelansel: K

michaelansel:
> I was trying to read up more on this too, and found this article useful: http://howtonode.org/understanding-process-next-tick

The bit about interleaving CPU intensive tasks is no longer true; you need to use `setImmediate` for that. This changed with 0.10.0.

Otherwise, yes, the article matches with my understanding and intent. If I get bored, I may take a pass through the codebase at some point and try to standardize things.

michaelansel:
> I saw some methods take a context object which has the listener and response. What if the robot.listenerMiddleware API took a context object instead of a listener and a response separately? I think that would help future proof the API a bit if we ever wanted to add anything else that's passed along.

I think that makes sense. When I did the refactor, I was trying to keep the already set API stable, but I'm game for updating it.

> I was also seeing a lot of interactions back and forth between robot to listener to robot. Do you think making a middleware class, or possibly a middleware stack class, would help with these interactions? That might simplify things, where a robot has a middleware stack, and listeners, and each listener just knows it has to go through the middleware stack.

Maybe? That was sort of the intent of the refactor (pulling all the execution logic into Robot). I'm not sure of exactly how to separate out the state here. We have a global Robot that contains most execution state, including a bunch of Listeners. Given that a Listener has a reference up to the containing Robot, it seems like it makes more sense to put middleware in the Robot. I could also see reason to put middleware in a separate object referenced by the Robot, however, I don't think this changes containment, just shuffles things around a little: you still have a Listener pointing back to the Robot for each execution of `call`.

An alternative architecture might be to have an executable Middleware object that gets injected into each Listener at creation time. You would then call the Middleware object to do all execution work (`middleware.execute context, next, done`). This introduces a change to the Listener constructor interface, but feels a little more like "proper" dependency injection. However, this doesn't significantly change the pattern, because we are already injecting the Robot and calling it. It would move execution logic out of the Robot though and into the Middleware class. Overall, I'm not sure if it buys us a lot, but it does feel a little cleaner/modular. It would probably be a little easier to test.

I'll take a shot at the alternative architecture and see what I come up with.

michaelansel:
And a pile of refactoring later... we have a separate Middleware class, clear separation with the Listener class, and a non-special API for Listener Middleware. Each commit should be fairly readable on its own and explain the progression.

Overall, I'm happy with all the changes, but I'm not completely sold on despecializing Listener Middleware. I think there is value in having `listener` and `response` exposed directly, but I can also see value in future proofing. My biggest concern is that "context" is very abstract, and we already have enough mind-bending required to understand middleware.

This also introduces an additional backwards-compatible API change: you can now specify an optional `middleware` argument to `Listener.call`. In the absence of a middleware stack to execute, `Listener.call` defaults to an empty stack.

michaelansel:
Because I'm weird, I did some playing around with git and discovered that if I rebase middleware on top of a pretend-merged #801, the diff turns out to be rather clean. It was nice self-validation that middleware is really just replacing `Robot.receive` and `Listener.call` and then adding a mountain of tests and documentation. Not sure if seeing that is valuable, but here's the branch diff if you are interested.

https://github.com/michaelansel/hubot/compare/listener-options...michaelansel:middleware2#diff-740956d8fdad73632c4937d631b7c79b

I'm starting to wonder if it might be useful to revisit #801 and merge it independently of #803, just for the sake of keeping this PR focused.

ambikads:
Is there an update on this issue?

michaelansel:
Not really. I don't think anyone has had time to think about it. I'm probably going to re-open #801 and get it up to date so that we can make *some* forward progress, but not sure when this will get merged...

michaelansel:
Okay, #801 has been reopened and updated with a couple bug fixes. Once that merges, it should be more clear what is changing with this PR and hopefully increase the likelihood of it getting merged.

bhuga:
A PR to a PR was weird and created a few additional (and useless) merge commits. I have a commit:

https://github.com/michaelansel/hubot/commit/f10ca4d37310d6a4aeb30de77b6df3fc620f3529

Let's use that for discussion? Or @michaelansel can just cherry-pick it and we can discuss it here.

The big thing that came to mind while I was wrestling with it was that the `-Sync` versions can wait for a second iteration.

michaelansel:
@bhuga I like it. I'll work on merging in your commit (with a few tweaks) and [my other documentation commit](https://github.com/michaelansel/hubot/compare/middleware...michaelansel:middleware-idea?name=middleware-idea) and we can continue the conversation here.

michaelansel:
(moved the non-listener middleware conversation over to michaelansel/hubot#2)

bhuga:
@michaelansel anything I can do to help this along? We're getting to the point internally where I want to make sure we're not diverging.

michaelansel:
Really just stuck on time right now. I haven't had time to think about pretty much anything Hubot related. The only outstanding thing I have is tweaking the interface. I'm thinking `robot.listenerMiddleware (context, next, done) ->` is the right way to go. With that changed, I'm game for merging this, adding receive middleware immediately afterwards, and punting on response middleware for a little while until I have time to dig into it more (because it can be hacked in using receive middleware). What do you think?

bhuga:
> I'm thinking robot.listenerMiddleware (context, next, done) -> is the right way to go

I think that's fine too.

> With that changed, I'm game for merging this, adding receive middleware immediately afterwards, and punting on response middleware for a little while until I have time to dig into it more (because it can be hacked in using receive middleware). What do you think?

That sounds solid. I'm a little wary diving into this today, especially with the docs in a separate pull. But if we land this, it'll be much easier for me to jump in and help with receive middleware directly.

michaelansel:
Okay, API updated. I read through the docs in michaelansel/hubot#2 and it looks like all that is left is either not yet implemented or already covered. Any other thoughts on this before I do a final review pass? I'll see if I can find someone new to give everything a read through...

bhuga:
> Any other thoughts on this before I do a final review pass? I'll see if I can find someone new to give everything a read through...

This sounds great to me. Excited to get my hands on this!

michaelansel:
While I'm getting addition review from coworkers, @technicalpickles do you have any reservations about merging this?

docs/scripting.md
```diff
@@ -592,7 +592,7 @@ In addition to a regular expression and callback, the `hear` and `respond` funct
 
 The most important and most common metadata key is `id`. Every Listener should be given a unique name (options.id; defaults to `null`). Names should be scoped by module (e.g. 'my-module.my-listener'). These names allow other scripts to directly address individual listeners and extend them with additional functionality like authorization and rate limiting.
```
cycomachead: Reading through the documentation, I had some thoughts.
* I think it'd be nice to bold "every", as this tool will be much much more useful as external packages are updated.
* It seems like it would be a good idea to mention, that existing packages _will_ need to be updated to be able to take advantage of this. Maybe this particular document isn't the best location, but at least for a while, I would probably be helpful to point out that these features will require support from package authors.

michaelansel: Good point! Not really a part of middleware though. Would you mind putting together a quick PR, and I'll merge it separately?

cycomachead: Yeah, I can do that in a bit.

Thanks for all this hard work! It will make my hubot's maintenance muuuuuuuch better!

michaelansel:
I've done a pretty careful review of the docs. Everything looks in order. Time to push the big green button!

bhuga:
Awesomesauce. Happy almost-birthday, #803!

yes cookie

cycomachead:
Thanks again @michaelansel, and everyone else! :+1:

sdimkov:
OMG did this just got merged?  Great job everyone :)

PR #821, opened 2014-11-21
Hubot, conversationalist

cameronmcefee:
I love hubot, but I don't really feel like I have a relationship with him. I'm sure, with all the things he's seen on the internet he has stories to tell. I'm sure he wishes he could get to know me… to ask _me_ questions for a change… to judge my life choices rather than just me judging his.

demo

I took this project on mainly just because I wanted to see if I could do it. I'm fairly sure there's a better way to do this, but _it works_. If this is on the right track and can be made to work with some tweaking, I'd love to see it merged. I'm also happy to rewrite based on advice for a different approach if that's plausible. And if this piques someone's interest to it a better way themselves, I'm happy just seeing the feature make it whether it's me that creates it or not.

One change that probably needs to be made is to fix it so it _only_ works when you address hubot directly. Currently he just listens to anything you say while a conversation is active with him, which could lead to some issues if trying to talk to coworkers and hubot simultaneously. Think: accidentally deploying your lunch plans to production. I'm not sure the best way to do that, which is what leads me to believe there's a better way to do all of this.

Any pointers would be appreciated. Once I know whether this is likely to be merged or not, I'll happily write docs and tests where necessary.

If you'd like to test this out, here's a gist with some scripts I wrote to test things: https://gist.github.com/cameronmcefee/643c8824a1888c40542f

### DIFF
--- a/src/adapters/campfire.coffee
+++ b/src/adapters/campfire.coffee
@@ -1,9 +1,9 @@
 HTTPS          = require 'https'
 {EventEmitter} = require 'events'
 
-Robot                                                = require '../robot'
-Adapter                                              = require '../adapter'
-{TextMessage,EnterMessage,LeaveMessage,TopicMessage} = require '../message'
+Robot                                                                    = require '../robot'
+Adapter                                                                  = require '../adapter'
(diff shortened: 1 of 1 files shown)


technicalpickles:
https://github.com/lmarkus/hubot-conversation

Issue #728, opened 2014-07-02
Some HUBOT environment variables not namespaced properly. 


stresler:
The Express instantiation in robot.coffee sources a few variables from the environment that aren't prefaced with the HUBOT_ prefix. This can easily lead to clashes with any other nodejs apps trying to source environment variables. 

    269   setupExpress: ->
    270     user    = process.env.EXPRESS_USER
    271     pass    = process.env.EXPRESS_PASSWORD
    272     stat    = process.env.EXPRESS_STATIC

and then 

    291     try
    292       @server = app.listen(process.env.PORT || 8080, process.env.BIND_ADDRESS || '0.0.0.0')

I noticed then when I was working on https://github.com/github/hubot/issues/727 where I am adding HUBOT_EXPRESS_KEY and HUBOT_EXPRESS_CERT

Let me know if this is the way you want to proceed and I can roll it all up in another PR.  Also if you want backward compatibility for any reason (e.g. accept PORT, but log a message to migrate to HUBOT_PORT

michaelansel:
Now that #839 has been merged, are there other variables that need to be corrected, or can this be closed?

stresler:
Yes. Awesome.

PR #839, opened 2014-12-22
Add Support for EXPRESS_PORT Variable

todd:
As discussed in #728, this PR adds support for an `EXPRESS_PORT` variable to set the port for Express to listen on while also retaining support for the current `PORT` variable. `EXPRESS_PORT` will be favored over `PORT`. My only outstanding question is whether we should also add `EXPRESS_BIND_ADDRESS` - this hasn't been an issue I've run into, but it might be nice for the sake of consistency.

cc: @technicalpickles

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -270,7 +270,7 @@ class Robot
     pass    = process.env.EXPRESS_PASSWORD
     stat    = process.env.EXPRESS_STATIC
     port    = process.env.EXPRESS_PORT or process.env.PORT or 8080
-    address = process.env.BIND_ADDRESS or '0.0.0.0'
+    address = process.env.EXPRESS_BIND_ADDRESS or process.env.BIND_ADDRESS or '0.0.0.0'
 
     express = require 'express'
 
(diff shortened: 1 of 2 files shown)


src/robot.coffee
```diff
@@ -269,6 +269,8 @@ class Robot
     user    = process.env.EXPRESS_USER
     pass    = process.env.EXPRESS_PASSWORD
     stat    = process.env.EXPRESS_STATIC
+    port    = process.env.EXPRESS_PORT or process.env.PORT or 8080
+    address = process.env.BIND_ADDRESS or '0.0.0.0'
```
michaelansel: I don't see any reason not to do the same namespacing with BIND_ADDRESS. Leaving it as the lone non-namespaced variable for Express just seems strange.

todd: Yeah, that was more or less my thinking as well, but I'm waiting for official word from a maintainer that that's what they want.

technicalpickles: `BIND_ADDRESS` was added in https://github.com/github/hubot/pull/609 . I'd be fine namespacing it, as long as the old BIND_ADDRESS is used as well.

technicalpickles:
I think this should be good to merge, once EXPRESS_BIND_ADDRESS is supported as well.

todd:
@technicalpickles I went ahead and added that. Should be good to go.

todd:
@technicalpickles Any chance we can get this merged?

todd:
:heart:

PR #848, opened 2015-01-22
Stop listening to uncaughtException events on shutdown

aroben:
I noticed this when running tests for a Hubot instance that created a lot of `Robot` instances. Eventually Node warned that there were more than 10 listeners for the `uncaughtException` event.

/cc @technicalpickles


aaronstaves:
Similar to my pull request from a (long) while back - https://github.com/github/hubot/pull/676.  Either way would love to see this event get cleaned up

bouzuya:
:+1:

PR #850, opened 2015-01-24
Revert "Add a link to the Fleep adapter"

2250918808:


technicalpickles:
Is there a reason the fleep adapter should be removed?

PR #770, opened 2014-09-20
update coffee-script v1.8.x

shokai:
I want to use latest coffee-script npm.
I've tested all `scripts/*.coffee` in hubot template.

PR #863, opened 2015-02-12
Bumping to use CoffeeScript 1.9.0

silasb:
CoffeeScript 1.9.0 came out Jan 29, 2015.  We are using 1.6.3 which is from Jun 2, 2013 and is quite old.

Some new features that I would like to use while building scripts are the 

[Splat Expansions:](https://github.com/jashkenas/coffeescript/blob/1.9.0/test/arrays.coffee#L62-L71)

```coffeescript
    [.., rel] = link.match /rel="(\w+)"/
```

I can't currently do this with coffee-script 1.6.3.

This is my first fix for hubot, so I'm not 100% sure on the procedure.  I saw a test folder, but not sure how to interact with the tests.  The only test that I did run was my script and it worked.  I also ran `script/smoke-test` and nothing blew up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/github/hubot.git"
   },
   "dependencies": {
-    "coffee-script": "1.6.3",
+    "coffee-script": "1.9.0",
     "optparse": "1.0.4",
     "scoped-http-client": "0.10.0",
     "log": "1.4.0",


technicalpickles:
cc https://github.com/github/hubot/pull/770 which is bumping to coffeescript 1.9.x

technicalpickles:
In particular, my comment here: https://github.com/github/hubot/pull/770#issuecomment-60941070

> One thing I'm wondering about is if we can remove this as a direct dependency of hubot. Hubot, the library, needs coffeescript, but really it's the instance of a hubot that should be controlling what coffeescript is used. I think that's specifically what peerDependency is for.
>
> The hubot generator would need to be updated to install coffee script as well. A positive consequence of that I think would also being able to not require a global coffeescript install (ie update generator-hubot's bin/hubot to add a local coffee to the PATH).

technicalpickles:
Another thought is that hubot can specify the minimum version needed to work with the coffeescript features it uses, while leaving the generated hubot to specify a specific version to use.

mal:
Big +1

Building a new hubot at the moment and would love to be able to escape `1.6.3`.

technicalpickles:
Could you try merging in master please? There are now better unit tests in place. I merged master into a local copy of your branch, and I'm getting:

```
$ npm install
npm WARN package.json log@1.4.0 No repository field.
npm WARN package.json readline-history@1.2.0 No repository field.
npm WARN unmet dependency /Users/technicalpickles/github/hubot/node_modules/hubot-mock-adapter/node_modules/hubot requires coffee-script@'1.6.3' but will load
npm WARN unmet dependency /Users/technicalpickles/github/hubot/node_modules/coffee-script,
npm WARN unmet dependency which is version 1.9.0
```

Then when I try to run `script/test`:

```
/Users/technicalpickles/github/hubot/node_modules/coffee-script/lib/coffee-script/coffee-script.js:210
          throw new Error("Use CoffeeScript.register() or require the coffee-s
                ^
Error: Use CoffeeScript.register() or require the coffee-script/register module to require .coffee.md files.
  at Object._base.(anonymous function) [as .coffee] (/Users/technicalpickles/github/hubot/node_modules/coffee-script/lib/coffee-script/coffee-script.js:210:17)
  at Module.load (module.js:356:32)
  at Function.Module._load (module.js:312:12)
  at Module.require (module.js:364:17)
  at require (module.js:380:17)
  at /Users/technicalpickles/github/hubot/node_modules/mocha/lib/mocha.js:185:27
  at Array.forEach (native)
  at Mocha.loadFiles (/Users/technicalpickles/github/hubot/node_modules/mocha/lib/mocha.js:182:14)
  at Mocha.run (/Users/technicalpickles/github/hubot/node_modules/mocha/lib/mocha.js:394:31)
  at Object.<anonymous> (/Users/technicalpickles/github/hubot/node_modules/mocha/bin/_mocha:394:16)
  at Module._compile (module.js:456:26)
  at Object.Module._extensions..js (module.js:474:10)
  at Module.load (module.js:356:32)
  at Function.Module._load (module.js:312:12)
  at Function.Module.runMain (module.js:497:10)
  at startup (node.js:119:16)
  at node.js:901:3
```

technicalpickles:
I changed the line in `script/test` to this, and it worked:

```bash
mocha --require coffee-script/register --compilers coffee:coffee-script "$@"
```

mal:
I believe that will only work while the peer'd `coffee-script` is `1.7.0` or greater. If that's ok for test purposes, I'll make that change, if not I guess that bash file will need to do a bit of version detection.

technicalpickles:
@mal yeah, that's definitely tricky. I think ideally, we don't have an explicit coffee-script version, so the bots can specify what they need. Maybe something like ">=1.6.3 1.x"? Then, we might be able to configure travis-ci in order to test all the versions.

mal:
Also (as per your suggestion) would be good to move it to a peer dependency, although I'm then unsure of how predictable the peer's binary location would be; as when hubot itself was a dependency it would be something like `../.bin/coffee`, but when just standalone testing I'm unsure where it would live.

technicalpickles:
I read the docs a bit better, and the line I pasted above can be simplified to:

```
mocha --compilers coffee:coffee-script/register "$@"
```

Feasibly do something like this in in script/test to choose the right thing:

```
$ coffee -e "console.log require('coffee-script').VERSION"
1.9.0
```

technicalpickles:
I've been trying to think through how exactly we are using coffee-script in hubot. Basically:

* `bin/hubot` has a shebang of `/usr/bin/env coffee`
* `bin/hubot` requires stuff that is coffee, but at that point, coffee should already be registered

Then for generated bots:

* `package.json` only specifies hubot (no coffee-script)
* `bin/hubot` is a shell script that does
   * `npm install`
  * alter PATH to include "node_modules/.bin:node_modules/hubot/node_modules/.bin:$PATH"
* runs `node_modules/.bin/hubot`

The reason this works is that having 'hubot' as a dependency, drops it in `node_modules/hubot`, with the github/hubot `bin/hubot` `node_module/.bin/hubot`. At that point, it looks for coffee on the path, which it  is because it's in `bin:node_modules/hubot/node_modules/.bin` because it's a dependency of hubot.

My understanding is that peerDependencies don't necessarily install something, but just indicate that they can use a peer if it's available.

Use cases we'll need to make sure work:

* newly generated bots
* bots previously generated

technicalpickles:
Ideally, I'd like to avoid having to have to bump and release hubot each time there's a new coffee-script. For one, that's means people are blocked on me reviewing and updating. And of more concern to me, is I'm worried of having coffee-script incompatabilities creep in, so one version of hubot works and another doesn't.

technicalpickles:
In other news, I just came across this: http://blog.davidbanham.com/articles/coffeescript-psa/

technicalpickles:
After digging into how coffee gets used, I had an idea. In an ex bot, you should be able to run `npm install --save coffee-script`. That'd create `node_modules/.bin/coffee`, and with the way that `bin/hubot` adjusts the PATH, that newer coffee executable would come first.

I tested this with a script like:

```coffee
coffee = require('coffee-script')

module.exports = (robot) ->
  robot.respond /coff(ee+)/, (msg) ->
    msg.send ":coffee: #{coffee.VERSION}"
```

And it works:

```
Hubot> Hubot coffee
:coffee: 1.9.1
```

@silasb could you try this idea and see if it works for you?

technicalpickles:
@silasb actually, looking at the test you linked to, the particular test dates back to 2011.  Check out http://coffeescript.org/#splats ... basically, I don't think the initial example you gave is actually using splats correctly.

mal:
@technicalpickles nice find on the fact that it uses the coffee in the hubot-app dependencies, this has gotten all my `>= 1.7` scripts working as they should :grinning:

In fact, this discovery is even better because now that I know this, simply adding the required version of coffee-script as a peerDependency to my script module is enough to get it working. Which is perhaps even better than having to specify the coffee version in the hubot app itself. This way scripts can all give their version constraints and we can let `npm` warn users about installed scripts that have conflicting version constraints, and all the while hubot can keep doing it's thing with `1.6.3` (until it can be bumped for real real).

silasb:
@technicalpickles Sorry, haven't had a chance to try this again.  I'll have some time 7-9 PM EST to try again.

silasb:
@technicalpickles that worked fine.

```
silas ~/hubot/node_modules/hubot master ❯ ./node_modules/.bin/coffee -v
CoffeeScript version 1.6.3
```

```
silas ~/hubot master ❯ ./node_modules/.bin/coffee -v
CoffeeScript version 1.9.1
```

Since this solves my problem, I'll close the bug.

PR #861, opened 2015-02-08
Rename LICENSE to LICENSE.md

SuriyaaKudoIsc:
Hi *guys*, :smile:

I renamed the **Licence file** to a `.md` file. :name_badge:
Now it looks better and similar like the `README.md` file.

Yours,
[**SuriyaaKudo**](https://github.com/SuriyaaKudoIsc) :octocat:

PR #860, opened 2015-02-08
Update date License

danilovaz:
Update year License and add License link in README.md

PR #870, opened 2015-02-20
Copyright update

technicalpickles:
This combines the ideas of https://github.com/github/hubot/pull/860 and https://github.com/github/hubot/pull/861 . I also wanted to reduce the number of places that would need to be updated yearly, so I removed mention of copyright in README.md and moved it to LICENSE.md.


LICENSE.md
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013 GitHub Inc.
```
technicalpickles: I did a `git log on README` to find the original year was 2011 cc b253e94e051c017eb7c8c0101c30a24f0851a499

PR #871, opened 2015-02-24
Add Robot#eavesdrop

### LINKED ISSUE #866: Have hubot respond to everything EXCEPT robot.respond?
`robot.hear` is triggered for every message. And `robot.respond` is only triggered where the message starts with the robots name. Is there a way to trigger only if the message doesn't start with the robots name?

michaelansel:
Lots of new tests and refactoring to eventually add Robot#eavesdrop and fix #866.

Each commit should be able to stand alone (all tests passing), refactor commits don't touch code and tests at the same time, and new tests were added first to cover all areas being modified.

`TextListener` was removed in favor of a common pattern of putting all matcher logic in Robot instead of some in Robot and some in a subclass of Listener.

This also introduces Robot#listen as a new public method.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -4,12 +4,13 @@ Path           = require 'path'
 HttpClient     = require 'scoped-http-client'
 {EventEmitter} = require 'events'
 async          = require 'async'
+{inspect}      = require 'util'
 
 User = require './user'
 Brain = require './brain'
 Response = require './response'
 {Listener,TextListener} = require './listener'
-{EnterMessage,LeaveMessage,TopicMessage,CatchAllMessage} = require './message'
+{EnterMessage,LeaveMessage,TextMessage,TopicMessage,CatchAllMessage} = require './message'
 Middleware = require './middleware'
 
 HUBOT_DEFAULT_ADAPTERS = [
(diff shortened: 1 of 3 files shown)


src/robot.coffee
```diff
@@ -75,7 +88,37 @@ class Robot
...
+
+  # Public: Adds a Listener that attempts to match incoming messages based on
+  # a Regex, but rejects all messages that are addressed to the robot
+  # (i.e. anti-respond).
+  #
+  # regex    - A Regex that determines if the callback should be called.
+  # callback - A Function that is called with a Response object.
+  #
+  # Returns nothing.
+  eavesdrop: (regex, callback) ->
```
michaelansel: I'm not super happy with the amount of duplication between `eavesdrop` and `hear`. I'm open to suggestions for refactoring.

src/robot.coffee
```diff
@@ -67,6 +69,17 @@ class Robot
       @emit 'error', err
     process.on 'uncaughtException', @onUncaughtException
 
+  # Public: Adds a custom Listener with the provided matcher and callback
```
michaelansel: By making this public, it does expose `Listener.matcher` as a new API. The API seems fairly well defined, but I can pull back to private if so desired.

src/robot.coffee
```diff
@@ -78,7 +91,39 @@ class Robot
   #
   # Returns nothing.
   hear: (regex, options, callback) ->
```
michaelansel: This results in `regex` being hidden from view. It should probably be added to the Listener as an option or something...

michaelansel:
I'll gut and rewrite this when #986 merges

bkeepers:
> I'll gut and rewrite this when #986 merges

@michaelansel still planning on doing this?

michaelansel:
And a bit of copy-paste later, we are all back up to date!

Apparently we lost our CI, but tests are passing on my laptop.

Issue #826, opened 2014-12-05
shell adapter throwing errors when trying to save history


technicalpickles:
I tried out the new shell adapter in https://github.com/github/hubot/pull/800 in a non-test hubot, and seeing problems:

```
Hubot> hubot hel[Fri Dec 05 2014 18:19:08 GMT-0500 (EST)] ERROR Error: EBADF, bad file descriptor
  at Object.fs.writeSync (fs.js:429:18)
  at Interface.rl._addHistory (/Users/technicalpickles/github/hubot-classic/node_modules/hubot/node_modules/readline-history/readline-history.js:106:20)
  at Interface._line (readline.js:516:19)
  at Interface._ttyWrite (readline.js:736:14)
  at ReadStream.onkeypress (readline.js:97:10)
  at ReadStream.EventEmitter.emit (events.js:126:20)
  at emitKey (readline.js:1058:12)
  at ReadStream.onData (readline.js:807:7)
  at ReadStream.EventEmitter.emit (events.js:96:17)
  at TTY.onread (net.js:397:14)


```

And on a different run:

```
ERROR Error: UNKNOWN, unknown error
  at Object.fs.writeSync (fs.js:429:18)
  at Interface.rl._addHistory (/Users/technicalpickles/github/hubot-classic/node_modules/hubot/node_modules/readline-history/readline-history.js:106:20)
  at Interface._line (readline.js:516:19)
  at Interface._ttyWrite (readline.js:736:14)
  at ReadStream.onkeypress (readline.js:97:10)
  at ReadStream.EventEmitter.emit (events.js:126:20)
  at emitKey (readline.js:1058:12)
  at ReadStream.onData (readline.js:807:7)
  at ReadStream.EventEmitter.emit (events.js:96:17)
  at TTY.onread (net.js:397:14)
```

The .hubot_history file does exist, but it's empty.

Issue #845, opened 2015-01-06
Running hubot with grunt-shell throws readline_history error


robert-rally:
I'm using Grunt to launch hubot using a grunt-shell command. This used to work fine, but now I get the following error:

[Tue Jan 06 2015 11:00:41 GMT-0700 (MST)] ERROR TypeError: Cannot read property 'push' of undefined
  at /hubot-test/node_modules/hubot/node_modules/readline-history/readline-history.js:112:19
  at ReadStream.onend (/hubot-test/node_modules/hubot/node_modules/readline-history/readline-history.js:52:21)
  at ReadStream.emit (events.js:117:20)
  at _stream_readable.js:943:16
  at process._tickCallback (node.js:419:13)

technicalpickles:
@robert-rally I'm really familiar with grunt-shell. Would you be able to share a code-snippet with how you are launching it?

robert-rally:
## Gruntfile.coffee
```coffeescript
module.exports = (grunt) ->
  grunt.loadNpmTasks 'grunt-shell'
  grunt.registerTask 'jarvis', ['shell:jarvis']

  grunt.initConfig
    shell:
      jarvis:
        command: './bin/hubot --name jarvis'
```
## bin/hubot
```shell
#!/bin/sh
npm install
export PATH="node_modules/.bin:node_modules/hubot/node_modules/.bin:$PATH"
exec node_modules/.bin/hubot "$@"
```
## package.json
```javascript
{
  "dependencies": {
    "hubot": ">= 2.6.0 < 3.0.0",
    "hubot-scripts": ">= 2.5.0 < 3.0.0"
  },
  "devDependencies": {
    "grunt-shell": "~1.1.1"
  },
  "engines": {
    "node": ">= 0.10.x",
    "npm": ">= 2.0.x"
  }
}
```

viking:
This also happens if you run hubot with `node debug`.

benbalter:
I can recreate this as well. Hubot runs fine with when executed with `./bin/hubot` but errs out when executed with a Procfile of `web: bin/hubot` and then `foreman start`.

robert-rally:
The bug is in readline_history. It's a bad monkey-patch that assumes readLine.history is always initialized. Spoiler alert: it isn't.

This history feature provides dubious value and definitely breaks things. Is it possible to simply remove it from hubot?

technicalpickles:
> This history feature provides dubious value

[...]

PR #877, opened 2015-03-09
Reimplement shell adapter using cline and chalk, and fix persistent history

technicalpickles:
https://github.com/github/hubot/pull/800 introduced a persistent history for the shell adapter. The main reason for this was to make it a bit easier to iterate in development, but it has caused other problems, particularly it's 

As @robert-rally pointed out about readline-history:

> It's a bad monkey-patch that assumes readLine.history is always initialized. Spoiler alert: it isn't.

This should fix https://github.com/github/hubot/issues/826 and https://github.com/github/hubot/issues/845 . It replaces using readline and readline-history with [cline](https://github.com/kucoe/cline), because it has a lot better interface. In particular, it does have a hook to managing a history. I just had to write some code to load it initially, and to truncate it at the end.

I also replaced the hard-coded terminal color output with [chalk](https://github.com/sindresorhus/chalk), so it's a lot easier to know what it's even doing.

cc @benbalter @keithduncan @viking @robert-rally who have run into problems in one way or another

### DIFF
--- a/src/adapters/shell.coffee
+++ b/src/adapters/shell.coffee
@@ -1,14 +1,13 @@
-Readline = require 'readline-history'
-
+Cline = require('cline')
 Robot         = require '../robot'
 Adapter       = require '../adapter'
 {TextMessage} = require '../message'
-
 
 historySize = if process.env.HUBOT_SHELL_HISTSIZE?
                 parseInt(process.env.HUBOT_SHELL_HISTSIZE)
               else
                 1024
+
 
 class Shell extends Adapter
   send: (envelope, strings...) ->
@@ -16,7 +15,6 @@ class Shell extends Adapter
       console.log "\x1b[01;32m#{str}\x1b[0m" for str in strings
     else
       console.log "#{str}" for str in strings
-    @repl.prompt()
 
   emote: (envelope, strings...) ->
     @send envelope, "* #{str}" for str in strings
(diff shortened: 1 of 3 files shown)


src/adapters/shell.coffee
```diff
@@ -12,11 +15,7 @@ historySize = if process.env.HUBOT_SHELL_HISTSIZE?
 
 class Shell extends Adapter
   send: (envelope, strings...) ->
-    unless process.platform is 'win32'
-      console.log "\x1b[01;32m#{str}\x1b[0m" for str in strings
-    else
-      console.log "#{str}" for str in strings
-    @repl.prompt()
+    console.log chalk.green.bold("#{str}") for str in strings
```
keithduncan: :heart_eyes: very nice, does `chalk` handle the win32 case too?

src/adapters/shell.coffee
```diff
@@ -26,41 +25,71 @@ class Shell extends Adapter
...
+    @cli.command '*', (input) =>
+      userId = parseInt(process.env.HUBOT_SHELL_USER_ID or '1')
+      userName = process.env.HUBOT_SHELL_USER_NAME or 'Shell'
+      user = @robot.brain.userForId userId, name: userName, room: 'Shell'
+      @receive new TextMessage user, input, 'messageId'
+
+    @cli.command 'history', () =>
+      console.log item for item in @cli.history()
+
+    @cli.on 'history', (item) =>
```
keithduncan: :+1:

src/adapters/shell.coffee
```diff
@@ -26,41 +25,71 @@ class Shell extends Adapter
...
+      if item.length > 0 and item isnt 'exit' and item isnt 'history'
+        fs.appendFile '.hubot_history', "#{item}\n", (err) =>
```
keithduncan: Could we de-duplicate `.hubot_history` into a constant for the below `close` event handler to reference too?

src/adapters/shell.coffee
```diff
@@ -26,41 +25,71 @@ class Shell extends Adapter
...
+        startIndex = history.length - historySize
+        history = history.reverse().splice(startIndex, historySize)
+
+        outstream = fs.createWriteStream('.hubot_history')
+        # >= node 0.10
+        outstream.on 'finish', () =>
+          @shutdown()
+
+        for item in history
+          outstream.write "#{item}\n"
```
keithduncan: I’d use `outstream.write history.join("\n")` to write more at once ad avoid having to loop ourselves

src/adapters/shell.coffee
```diff
@@ -26,41 +25,71 @@ class Shell extends Adapter
...
+
+        # < node 0.10
+        outstream.end () =>
+          @shutdown()
+       else
+         @shutdown()
+
+  loadHistory: (callback) ->
+    fs.exists '.hubot_history', (exists) ->
```
keithduncan: Though marginal this is a time-of-check time-of-use race, the file could technically disappear between the initial check and the read. Would it be possible to use the read and error handle `ENOENT` (not sure what we get from `fs.createReadStream` if it doesn’t exist)

src/adapters/shell.coffee
```diff
@@ -26,41 +25,71 @@ class Shell extends Adapter
...
+        for item in history
+          outstream.write "#{item}\n"
+
+        # < node 0.10
+        outstream.end () =>
+          @shutdown()
+       else
+         @shutdown()
+
+  loadHistory: (callback) ->
```
keithduncan: Could we add docs for this method and the types of callback and it’s argument?

keithduncan:
This looks great, I tested it out and it works nicely :smile:

src/adapters/shell.coffee
technicalpickles: It should, yeah. It uses the `supports-color` module for determining if the terminal supports color or not.

src/adapters/shell.coffee
technicalpickles: Done.

src/adapters/shell.coffee
technicalpickles: I was originally thinking to not join it to avoid keeping it in memory. It shouldn't be a big deal though, especially since it's closing.

technicalpickles: Also, joining on "\n" is slightly different than writing one line out at a time (ie there's not a trailing newline).

src/adapters/shell.coffee
technicalpickles: I just tried doing this, and didn't have much luck, ie hubot was logging the error w/o a stack trace and hanging. I think it's going to be okay as is for now.

technicalpickles:
>  I just noticed this creates the history file in CWD, it might be nice to make this create the file in ../ relative to script/server so that it’s always co-located with the hubot instance being run.

cc https://github.com/github/hubot/pull/726 for this

src/adapters/shell.coffee
keithduncan: > I was originally thinking to not join it to avoid keeping it in memory.

That’s legit :+1:

> (ie there's not a trailing newline).

`Array#join` doesn’t add a trailing newline, the separator is just added inbetween the items:

```
coffee> [ 'a', 'b' ].join("\n")
'a\nb'
```

PR #881, opened 2015-03-13
Update docs for rendering on hubot.github.com

technicalpickles:
We recently updated hubot.github.com, but the docs there were duplicated from this docs/ directory.

This makes things work on hubot.github.com and be linkable, but they won't be linkable when viewing on github.com.

cc @gjtorikian who I was talking to this about


docs/adapters.md
```diff
@@ -1,4 +1,7 @@
-# Hubot Adapters
+---
+permalink: /docs/adapters/index.html
+layout: docs
```
parkr: You might find [YAML Front Matter Defaults](http://jekyllrb.com/docs/configuration/#front-matter-defaults) helpful in reducing the repetition here.

parkr:
Why specify an explicit permalink? It looks like you're using the standard structure. It links to the folder instead of the index.html inside.

docs/adapters.md
technicalpickles: I tried that in hubot.github.com's _config.yml, but it didn't seem to take effect:

```yaml

defaults:
  -
    scope:
      path: "docs"
      type: "submodules"
    values:
      layout: "docs"
```

technicalpickles:
@parkr it's because this is being included as a git submodule in another gh-pages repo. In that repo, it's _submodules/hubot/docs/index.html, for example, and without the permalink, it was generating to submodules/hubot/docs/index instead of docs/index.html.

docs/adapters.md
parkr: the type of these are `pages` actually :)

parkr: oh I see!

parkr: try removing `path?

parkr:
> @parkr it's because this is being included as a git submodule in another gh-pages repo. In that repo, it's _submodules/hubot/docs/index.html, for example, and without the permalink, it was generating to submodules/hubot/docs/index instead of docs/index.html.

[you can change the permalink for collections!](https://github.com/jekyll/jekyll/blob/v2.4.0/lib/jekyll/collection.rb#L154)

```
collections:
  submodules:
    permalink: /:path/
```

PR #882, opened 2015-03-13
Fix links for viewing on hubot.github.com

technicalpickles:


docs/adapters/campfire.md
```diff
@@ -16,7 +16,7 @@ Next, you will need to create a user on your Campfire account for your Hubot,
 then give it access so it can join to your rooms. You will need to create a room
 if you haven't already.
 
-Hubot defaults to using its [shell](shell.md), so to use Campfire instead, you
+Hubot defaults to using its [shell](/docs/adapters/shell), so to use Campfire instead, you
```
parkr: If you're using folders (e.g. `/docs/adapters/shell/index.html`), you'll see a 301 with these links and Google, when following them, will dock this page.

technicalpickles: @parkr fixed that, thanks!

Issue #731, opened 2014-07-11
Using hubot behind a proxy server


juanpaulo:
Use hubot behind http proxies. Possibly replace current http client to request (https://github.com/mikeal/request).
Consider backward compatibility for scripts, etc.
Consider how to pass proxy info as well.

technicalpickles:
I've never worked behind a HTTP proxy before, so it's going to be hard for me to fix this. If I knew what type of proxies are being used, could probably work out a vagrantfile to bring up a development environment behind one.

nottinhill:
When will this be possible? I am sorry to hear that devs in here actually have never worked behind a proxy - it is almost unbelievable - am I the only one employed here?

technicalpickles:
@nottinhill it is a pretty low priority for my hubot needs so I'm not going to get to it anytime soon.

I'm happy to review and merge pull requests to support it though!

michaelansel:
@nottinhill We have pretty hard proxy requirements and using the above gist has been enough to get everything working. I'll look at making the injection method less of a hack, but the gist should be enough to allow you to move forward.

nottinhill:
Sorry to say - I long moved toward different kind of Software with Proxy feature build in.

MitchK:
@nottinhill Check my fix for this: https://github.com/github/hubot/issues/287#issuecomment-78453521

PR #884, opened 2015-03-13
Add support for global http options

### LINKED ISSUE #287: Hubot behind a proxy
I'm behind a proxy, and while `hubot ping` works just fine, `math me 2+2` doesn't work. In fact, any step that uses a `get()` in the script just doesn't work at all. How do I either:
1. Tell Hubot to use a proxy when doing a `.get()`?
2. Get some additional debugging out of Hubot so that I can diagnose why things aren't working?

Thanks!

michaelansel:
Primary purpose is to support global proxy settings in corporate environments

Closes #287 and #731

### DIFF
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -72,7 +72,8 @@ In many corporate environments, a web proxy is required to access the Internet a
 Due to the way node.js handles HTTP and HTTPS requests, you need to specify a different Agent for each protocol. ScopedHTTPClient will then automatically choose the right ProxyAgent for each request.
 
 ```coffeescript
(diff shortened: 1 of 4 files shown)


michaelansel:
Soft dependency on technoweenie/node-scoped-http-client#31 (code will work, but docs are a lie until that merges)

michaelansel:
Dependent commit merged, so this should be good to go now

michaelansel:
Dependent commit merged; just waiting on a release including the new behavior

technicalpickles:
0.11.0 is released now

Issue #786, opened 2014-10-10
Update documentation and examples to use `res` instead of `msg`


technicalpickles:
Whenever you see `robot.respond /something/, (msg) ->`, you'd think that msg is a `Message`. But it's not, it's actually a `Response`.

It's been proposed to update all existing examples and documentation to use `res` instead cc https://github.com/github/hubot/pull/764#issuecomment-56042492

It would be worth taking a pass through hubot-scripts as well, in order to remove prior art so people creating new scripts only see the new way of doing things.

technicalpickles:
One downside of this is that most scripts everywhere refer to `msg` instead of `res`, since they followed what documentation we do have :sweat:

Taytay:
For what it's worth, I am in strong favor of this change. Once I realized that `msg` was actually a `response` object, I started writing it as `response` in my scripts, and felt things were much clearer for it. It's true that legacy scripts will have this issue, but that's a small price to pay for future clarity.

geoffreyanderson:
@technicalpickles is this issue good to close out now that you've merged the associated PRs, or is there some other follow up work that's needed (e.g. tweaking some basic scripts to use `res` instead of `msg`)?

technicalpickles:
@geoffreyanderson yep!

PR #888, opened 2015-03-13
Update documentation to use `res` instead of `msg`

geoffreyanderson:


geoffreyanderson:
This and https://github.com/github/generator-hubot/pull/36 are targeted at answering #786.

technicalpickles:
Thanks for taking a stab at this! I'm surprised there was only one reference to `Response` as `res` in the code base.

This should be good to merge, I'm just little it sit out a little in case we can think of a reason this is a bad idea in https://github.com/github/hubot/issues/786 :grin:

geoffreyanderson:
Sounds good to me :+1:

technicalpickles:
Finally ready for this :grin: Doing a review, found just one `msg` references:

* https://github.com/github/hubot/blob/master/src/robot.coffee#L64-L65 (invokeErrorHandlers's definition is correctly using res though)

geoffreyanderson:
yay! :beers:

MrSaints:
A lot of scripts (past and present) uses `res` when making HTTP requests, e.g. https://github.com/hubot-scripts/hubot-google-translate/blob/master/src/google-translate.coffee#L103

```coffee
get() (err, res, body) ->
```

Anyone working with HTTP requests should keep this in mind otherwise you may end up with `undefined` errors. I thought I should raise this in case anyone is stuck in a pickle.

technicalpickles:
I've run into it with our own hubot actually, so yeah, kind of a problem :sweat:

I think it will only affect new scripts being written though, as it's just a variable name change local to the script. I've taken to using httpRes for those, but not sure if that is best or not yet.

PR #893, opened 2015-03-16
Update adapters.md

rwitchell:
fixed links


technicalpickles:
This is actually intentional. The documentation is now viewable on https://hubot.github.com/ (it's embeded as a submodule) , but there's not good way to have them viewable both from github.com/github/hubot and on the website that I've found yet :sweat:

technicalpickles:

Issue #828, opened 2014-12-08
Start up warnings


mal:
Apologies if this has already been reported, I did have a good search but couldn't find anything.

I encountered three minor issues during startup:

 - npm warning about hubot-maps
 - connect deprecation warnings
 - Redis info log gets printed after the first input prompt

```sh
$ bin/hubot                                                                                           
npm WARN package.json hubot-maps@0.0.0 No repository field.
connect deprecated multipart: use parser (multiparty, busboy, formidable) npm module instead node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/bodyParser.js:56:20
connect deprecated limit: Restrict request size at location of read node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/multipart.js:86:15
Hubot> [Mon Dec 08 2014 11:36:47 GMT+0000 (GMT)] INFO Using default redis on localhost:6379
```

mal:
Thanks @technicalpickles :+1:

PR #899, opened 2015-03-18
Eliminate "connect deprecated multipart" and "connect deprecated limit:" warnings

Taytay:
This replaces the deprecated express.bodyParser() with express.urlencoded() and express.json()

This gets rid of the two warnings reported in #828:
```
connect deprecated multipart: use parser (multiparty, busboy, formidable) npm module instead node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/bodyParser.js:56:20
connect deprecated limit: Restrict request size at location of read node_modules/hubot/node_modules/express/node_modules/connect/lib/middleware/multipart.js:86:15
```

(That issue was closed, but IMO the problem remains: These warnings are red herrings that serve to scare new users when something isn't working.)

express.bodyParser() is deprecated in favor of only using the parts you need.
I've replaced it with only the parsing bits that I believe Hubot is using: url encoded and json.
If Hubot does in fact need to parse multipart uploads, this patch will not work.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -288,7 +288,13 @@ class Robot
 
     app.use express.basicAuth user, pass if user and pass
     app.use express.query()
-    app.use express.bodyParser()
+
+    # Removing bodyParser since it's deprecated
+    # Note that Hubot no longer supports multipart uploads due to no longer using express.bodyParser
+    # In order to put support back for that, we would need to use one of: formidable, multiparty, busboy, multer, etc.
+    app.use express.urlencoded()
+    app.use express.json()
+
     app.use express.static stat if stat
 
     try


cbelsole:
:+1:

technicalpickles:
Thanks for looking into this! They have definitely been annoying.

> If Hubot does in fact need to parse multipart uploads, this patch will not work.

I think it's less a matter of hubot needing to parse multipart uploads (it itself doesn't), and more if people that have built on hubot are using scripts that do.

I think to maintain compatibility (and thus not need to change the major version number), it'd be worth considering adding something back in that handles multipart. It'd be worth doing a cursory check through the open source hubot scripts to find any that depend on this behavior.

Taytay:
Thanks for the input! It might be a couple of weeks before I can look into
multipart replacements. There are definitely a lot of options out there,
and I'm not sure what is drop-in simple. Suggestions welcome (I'm a
Node/Express newbie). If there is a chance that others are using them, I
agree I don't want to remove the functionality.

src/robot.coffee
```diff
@@ -288,7 +288,13 @@ class Robot
 
     app.use express.basicAuth user, pass if user and pass
     app.use express.query()
-    app.use express.bodyParser()
+
+    # Removing bodyParser since it's deprecated
+    # Note that Hubot no longer supports multipart uploads due to no longer using express.bodyParser
+    # In order to put support back for that, we would need to use one of: formidable, multiparty, busboy, multer, etc.
+    app.use express.urlencoded()
```
technicalpickles: Reading over http://www.senchalabs.org/connect/bodyParser.html, it says it's equivilant to:

```javascript
app.use(connect.json());
app.use(connect.urlencoded());
app.use(connect.multipart());
```

That makes me think the order is significant, so these should probably be swapped.

technicalpickles:
Did a little more digging around, and came across this: http://www.senchalabs.org/connect/multipart.html which talks about the deprecation. So does https://github.com/senchalabs/connect/wiki/Connect-3.0

I think we want to fix the deprecation warning without breaking compatibility (for now, longer term https://github.com/github/hubot/pull/772 is probably better for this kind of thing).

I'm having a hard time determining which would be best. https://github.com/andrewrk/connect-multiparty seems the simplest, but it does have a caveat like:

> I actually recommend against using this module. It's cleaner to use the multiparty API directly.
>
>This middleware will create temp files on your server and never clean them up. Thus you should not add this middleware to all routes; only to the ones in which you want to accept uploads. And in these endpoints, be sure to delete all temp files, even the ones that you don't use.

I looked at the others, and it looks like the API is a bit different. I was wondering what expressjs itself was doing, and turns out it's using this: https://github.com/expressjs/body-parser but note:

> This does not handle multipart bodies, due to their complex and typically large nature. 

I think this means needing do a major version bump, but... that doesn't seem great for this kind of change just to fix deprecation warnings. Could maybe get away with a minor version bump, but that seems like it's not really following semantic versioning because somethings _would_ be incompatible (ie hubot and hubot scripts out there using multipart).

technicalpickles:
> This middleware will create temp files on your server and never clean them up. Thus you should not add this middleware to all routes; only to the ones in which you want to accept uploads. And in these endpoints, be sure to delete all temp files, even the ones that you don't use.

I read https://github.com/andrewrk/connect-multiparty/blob/master/index.js a little more closely, and it's basically the same code as http://www.senchalabs.org/connect/multipart.html . It actually parse the form data unless it's a POST and is actually form-data, so it won't leave files around unless you are actually uploading files, so it's no different than the previous version.

Like I said earlier, we'll want to put this control more in the hands of user, but for now I think it'd be okay to go with connect-multiparty.

technicalpickles:
I'm trying to get some tests working testing multipart uploads, and it's a bit tricky.

Digging into https://github.com/andrewrk/node-multiparty/blob/master/index.js vs http://www.senchalabs.org/connect/multipart.html, it looks like the deprecated connect multipart also uses a limit middleware, which is being removed in 4.0 since none of the parses will use it anymore. The wiki says:

> Remove limit middleware. No longer used by any of the parsers. It doesn't work properly for streams2. If you are looking for limit middleware but are not using any of the body parsing middleware, you may be interested in [raw-body](https://github.com/stream-utils/raw-body).

technicalpickles:
I was able to put together an example script (based on the scripting.md):

```coffeescript
fs = require 'fs'

module.exports = (robot) ->
  robot.router.post '/hubot/chatsecrets/:room', (req, res) ->
    room   = req.params.room


    if req?.files?.secret
      file = req.files.secret
      fs.readFile file.path, (err, data) ->
        if err?
          robot.emit 'error', err
          res.send 'FAIL'
          return

        robot.messageRoom room, "I have a secret: #{data}"
        res.send 'OK'
        fs.unlink(file.path)
    else if req.body?
      data = if req.body.payload?
        JSON.parse req.body.payload
      else
        req.body
      secret = data.secret
      robot.messageRoom room, "I have a secret: #{secret}"
      res.send 'OK'
```

You can test this with:

```
echo "C-TECH Astronomy" > secret
curl -v -include --form secret=@secret http://127.0.0.1:8080/hubot/chatsecrets/Shell
```

technicalpickles:
I have a couple of changes on top of this, so I'm going to merge this branch as is, and have a quick followup PR.

PR #900, opened 2015-03-18
Improve error messages when a script or adapter can't be loaded.

Taytay:
Previously, if an adapter failed to load due to a syntax error, the location of the syntax error in the file was not specified. (This made for difficult debugging when writing a new adapter)
Now, the message is a descriptive:
`ERROR Unable to load adapter hubot-github-adapter(17,82) : Error: unexpected ,`
Line numbers are similarly made available when loading scripts (if available).

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -223,8 +223,17 @@ class Robot
           @logger.warning "Expected #{full} to assign a function to module.exports, got #{typeof script}"
 
       catch error
-        @logger.error "Unable to load #{full}: #{error.stack}"
+        @logger.error "Unable to load #{@getSyntaxErrorString(full, error)}"
         process.exit(1)
+
+
+  getSyntaxErrorString: (filename, syntaxError) ->
+    #(5,1):
+    message = "#{filename}"
+    if (syntaxError.location)
+      message += "(#{syntaxError.location.first_line},#{syntaxError.location.first_column})"
+    message += " : Error: #{syntaxError.message}"
+    return message
 
   # Public: Loads every script in the given path.
   #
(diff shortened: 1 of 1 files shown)


technicalpickles:
This looks pretty useful! The first questions that come to mind are:

* how does this fail for non-syntax errors? For example, if you try to access a property of undefined (ie process.env.NONEXISTENT.length when it's not set)
* how does the output compare to running `coffee -c path/to/thing.coffee`? That is the test I usually use when a syntax failure happens

Taytay:
1) Non syntax errors:
When adding that non-syntax error near the top of my adapter:
`[Wed Mar 18 2015 10:45:50 GMT+0000 (GMT)] ERROR Unable to load adapter hubot-github-adapter : Error: Cannot read property 'length' of undefined`
So, not terribly useful.

When adding that line to a script:
`[Wed Mar 18 2015 10:46:53 GMT+0000 (GMT)] ERROR Unable to load /Users/taylorbrown/projects/YNAB/ybot/scripts/github-pr-helper : Error: Cannot read property 'length' of undefined`

Also not terribly useful. :)

So, I think echoing the stack is going to be helpful in those cases (which I believe I removed) - I'll have to look into that.

1) Compared to `coffee`:
Hubot output
```
[Wed Mar 18 2015 10:49:53 GMT+0000 (GMT)] ERROR Unable to load adapter hubot-github-adapter(17,82) : Error: unexpected ,
```
coffee output:
```
/Users/taylorbrown/projects/YNAB/ybot/node_modules/hubot-github-adapter/src/hubot-github-adapter.coffee:18:83: error: unexpected ,
    return console.error "No Github token provided to Hubot" unless @options.token,,,
                                                                                  ^
```

technicalpickles:
Thanks for looking into that!

> So, I think echoing the stack is going to be helpful in those cases (which I believe I removed) - I'll have to look into that.

That sounds like a good plan. I think if you can detect it's a syntax error, you can omit logging the stack trace.

bkeepers:
@Taytay thanks for the pull request. It looks like there were still some changes to be made here. I'm going to go ahead and close this, but we'd be happy to reopen if you want to make the suggested changes.

Issue #880, opened 2015-03-12
Setting brain value on 'loaded' - Maximum call stack exceeded


victorhooi:
It's not mentioned in the official scripting documentation (https://github.com/github/hubot/blob/master/docs/scripting.md), however, it is mentioned elsewhere (http://www.icicletech.com/blog/automate-your-development-activities-with-hubot and http://thejacklawson.com/how-to-hubot/#/21) that you can use `robot.brain.on 'loaded', ->` to set variables when the Redis brain is ready.

I'm using the following snippet:

```
module.exports = (robot) ->

  robot.brain.on 'loaded', ->
    robot.brain.set 'aliases', {}

  robot.hear /alias set (\S*) (.*)/i, (msg) ->
    robot.brain.set aliases[msg.match[1]], msg.match[2]
    console.log(JSON.stringify(robot.brain.aliases))
    msg.reply("Alias set for " + msg.match[1])
```

and I get a `Maximum call stack exceeded` error when I try to start up Hubot:

```
[Fri Mar 13 2015 07:21:17 GMT+1100 (AEDT)] ERROR RangeError: Maximum call stack size exceeded
  at extend (/Users/victorhooi/foobar/foobar/foobar/node_modules/hubot/src/brain.coffee:155:3, <js>:149:67)
  at Brain.set (/Users/victorhooi/foobar/foobar/foobar/node_modules/hubot/src/brain.coffee:30:5, <js>:34:7)
  at Brain.<anonymous> (/Users/victorhooi/foobar/foobar/foobar/scripts/example.coffee:15:5, <js>:4:26)
  at Brain.emit (events.js:107:17)
  at Brain.set (/Users/victorhooi/foobar/foobar/foobar/node_modules/hubot/src/brain.coffee:31:5, <js>:35:12)
  at Brain.<anonymous> (/Users/victorhooi/foobar/foobar/foobar/scripts/example.coffee:15:5, <js>:4:26)
  at Brain.emit (events.js:107:17)
  at Brain.set (/Users/victorhooi/foobar/foobar/foobar/node_modules/hubot/src/brain.coffee:31:5, <js>:35:12)
  at Brain.<anonymous> (/Users/victorhooi/foobar/foobar/foobar/scripts/example.coffee:15:5, <js>:4:26)
  at Brain.emit (events.js:107:17)
  at Brain.set (/Users/victorhooi/foobar/foobar/foobar/node_modules/hubot/src/brain.coffee:31:5, <js>:35:12)
[...]

PR #903, opened 2015-03-22
Add more automated tests

michaelansel:
Another batch of tests covering:
* Robot: listener registration (complete)
* Robot: script loading (incomplete, just `loadFile`)
* Brain: everything (complete)

### DIFF
--- a/test/brain_test.coffee
+++ b/test/brain_test.coffee
@@ -1,0 +1,253 @@
+# Assertions and Stubbing
+chai = require 'chai'
+sinon = require 'sinon'
+chai.use require 'sinon-chai'
+
+{ expect } = chai
+
+# Hubot classes
+Brain = require '../src/brain.coffee'
+User = require '../src/user.coffee'
+
+describe 'Brain', ->
+  beforeEach ->
+    @clock = sinon.useFakeTimers()
+    @mockRobot =
+      emit: ->
+      on: ->
+
+    # This *should* be callsArgAsync to match the 'on' API, but that makes
+    # the tests more complicated and seems irrelevant.
+    sinon.stub(@mockRobot, 'on').withArgs('running').callsArg(1)
+
+    @brain = new Brain @mockRobot
+
+    @user1 = @brain.userForId '1', name: 'Guy One'
+    @user2 = @brain.userForId '2', name: 'Guy One Two'
+    @user3 = @brain.userForId '3', name: 'Girl Three'
+
+  afterEach ->
+    @clock.restore()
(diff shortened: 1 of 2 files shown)


test/brain_test.coffee
```diff
@@ -0,0 +1,253 @@
...
+
+    describe '#userForId', ->
+      it 'returns the user object', ->
+        expect(@brain.userForId(1)).to.equal(@user1)
+
+      it 'does an exact match', ->
+        user4 = @brain.userForId('FOUR')
+        expect(@brain.userForId('four')).to.not.equal(user4)
+
+      # Cannot understand why this behavior is needed, but adding a test case
```
michaelansel: I think this [can be removed](https://github.com/github/hubot/pull/670#issuecomment-74808011), but I figure it is better to have the tests and then remove them later.

test/brain_test.coffee
```diff
@@ -0,0 +1,253 @@
...
+        @brain.set
+          key2: 'val2'
+          key3: 'val2'
+
+        expect(@brain.data._private).to.deep.equal
+          key1: 'val1'
+          key2: 'val2'
+          key3: 'val2'
+
+      # Unable to understand why this behavior is needed, but adding a test
```
michaelansel: I think this is the cause for #880, but I figure it is better to have tests and remove them than to have incomplete coverage.

technicalpickles: re the comment, This is event is used to indicate when the brain has been 'loaded', so scripts can have a `robot.brain.on 'loaded'` to initialize some defaults.

michaelansel: Best I can tell, that is what `mergeData` is for. I did some spot checking around and every Brain persistence implementation I can find uses `mergeData` to inject data instead of using `set`.

Also, you can't initialize defaults on `loaded` because calling `set` from `loaded` triggers another `loaded` event (infinite loop). This behavior causes *every* `set` to emit `loaded`, which is confusing and, I'm pretty sure, wrong.

michaelansel:
@technicalpickles any objection to merging these tests?

technicalpickles:
@michaelansel no objections, but I would like to do a pass through to make sure I can understand what's going on :grin:

michaelansel:
Been a couple months. Poking this again...

bkeepers:
I just merged master locally and they all still pass. Since they are tests and won't affect any APIs, they can easily be change later. I vote :+1:.

sdimkov:
:+1:

PR #909, opened 2015-03-26
Update deploying docs with extension

andela-osogunle:
Clicking these links on master are broken, but the docs they meant to refer to are in /docs/deploying but only the extensions (.md) were missing. They are fixed now, and the links would work routing to the appropriate doc sheet.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -3,6 +3,6 @@ permalink: /docs/deploying/index.html
 layout: docs
 ---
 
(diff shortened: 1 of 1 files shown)


technicalpickles:
These docs are used on hubot.github.com, and so we had to use like /docs/deploying, so this is intentional for now. It'd be nice to find a way to have them work correctly when browsing on github.com and rendered on the site though.

technicalpickles:

PR #910, opened 2015-03-26
Update deploying docs with md file extension

andela-osogunle:
Clicking these links on master are broken, but the docs they meant to refer to are in /docs/deploying but only the extensions (.md) were missing. They are fixed now, and the links would work routing to the appropriate doc sheet.


technicalpickles:

PR #911, opened 2015-03-27
Fixed deploying doc links for heroku, unix and windows

ravikiranj:
The deploying docs for Heroku, Unix and Windows were pointing to a folder instead of the markdown files. This pull request fixes them.


technicalpickles:
This is intentional. The documentation is used to render on hubot.github.com, and unfortunately there's not currently a good way to have them work on both hubot.github.com and when browsing on github.com/github/hubot

technicalpickles:

Issue #916, opened 2015-04-03
Fix links in docs to work while browsing on GitHub and on hubot.github.com


technicalpickles:
It happened quietly, but the docs/ directory is now used to power hubot.github.com. Part of that was changing links to point to like /docs/scripting/ instead of docs/scripting.md. On the downside, we now have broken links if you browse https://github.com/github/hubot/tree/master/docs . We can make sure the links all work inconsistently, but the documentation has lived in the repository for so long, people link to them directly.

Since then, we've had a ton of PRs to fix the link the link, and I have to explain why this is. I expect to see more of them :sweat: 

So, let's figure out a way to make them work in both. This is complicated by hubot.github.com being a separate private repo for historical reasons.

cc @jglovier @gjtorikian @parkr @benbalter since we've talked about this at one point or another

gjtorikian:
I can't imagine any clean workaround for this. In `hubot.github.com`, we have `script/bootstrap` call `git submodule update --init --recursive`. As part of that bootstrap process, we could rewrite links in the submodule that match the GitHub format into a path for Jekyll. We could then add [`ignore = dirty`](http://www.nils-haldenwang.de/frameworks-and-tools/git/how-to-ignore-changes-in-git-submodules) to ignore these changes in the submodule.

That's all I got. :disappointed: Maybe we can/should also import the content (sans history) into `gh-pages` instead of keeping it separate. :neckbeard:

technicalpickles:
Could we make the permalinks have a .md? That is, /docs/scripting.md instead of /docs/scripting/index.html like it is now. I wonder if that would be correctly served up as html or not once generated...

Or maybe /docs/scripting.md/index.html. That way, links to /docs/scripting.md would work, albeit with a redirect in there.

jglovier:
>Could we make the permalinks have a .md

I don't think that will work. Since those links are just href attributes in the HTML, Jekyll doesn't know to convert them to anything.

technicalpickles:
> Or maybe /docs/scripting.md/index.html

Surprisingly, it totally does. At least, in development it works.

technicalpickles:
Could really use some help on this. Seems like we get a PR to 'fix' individual pages every few days, and it's really annoying to close them without a path to resolution.

parkr:
[...]

PR #920, opened 2015-04-14
Fix link to developing adapters

caseymrm:
Figured I should fix it after the third time I clicked on it.


technicalpickles:
It's super confusing, but this is intentional so the links are viewable on hubot.github.com :sweat: 

Tracking fixing that in https://github.com/github/hubot/issues/916

PR #921, opened 2015-04-15
added robot.once function, wrapper for events.once

shokai:
I have added `robot.once` function.

[events.once](https://nodejs.org/api/events.html#events_emitter_once_event_listener) is useful for handling one time event.
It is available at [node v0.3.0](https://nodejs.org/docs/v0.3.0/api.html)

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -437,6 +437,17 @@ class Robot
   # Public: A wrapper around the EventEmitter API to make usage
   # semanticly better.
   #
+  # event    - The event name.
+  # listener - A Function that is called with the event parameter
+  #            when first event happens.
+  #
+  # Returns nothing.
+  once: (event, args...) ->
+    @events.once event, args...
+
+  # Public: A wrapper around the EventEmitter API to make usage
+  # semanticly better.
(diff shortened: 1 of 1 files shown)


michaelansel:
This looks good to me. @technicalpickles, do you think we should have test cases for simple wrappers like this? It seems unnecessary if we are just matching the underlying API.

sdimkov:
+1 on this

technicalpickles:
> A wrapper around the EventEmitter API to make usage semantically better.

As we add more of these, it makes me wonder if this is actually true. Is it any easier than using `robot.events.once`, `robots.events.on`, `robots.events.emit`, etc?

technicalpickles:
I think the only thing this is missing is some docs in `docs/scripting.md` in the Events section.

timkinnane:
Just pinging this issue because I don't think it's resolved as determined, re updating the docs. I also had wondered about the usage, because I tried using chai-eventemitter to test Robot events but it didn't work because of the missing once. Now I realise I could have tested `robot.events` - so I think some clarification in the docs would actually be useful.

PR #922, opened 2015-04-19
Add Brain#usersForAny(query)

sdimkov:
Matches search query against all properties of users.

Very useful as universal search via email/mobile/ext/etc.. when your Hubot
knows more about the users than just name and id.

### DIFF
--- a/src/brain.coffee
+++ b/src/brain.coffee
@@ -148,6 +148,21 @@ class Brain extends EventEmitter
 
     matchedUsers
 
+  # Public: Get all users that contain the provided query string
+  # in any of their properties.
+  #
+  # Returns an Array of User instances matching the search criteria.
+  usersForAny: (query) ->
+    query = query.toLowerCase()
+    matchedUsers = []
+    for key, user of (@data.users or {})
+      for property, value of user
+        if typeof value is 'string'
+          if value.toLowerCase().indexOf(query) isnt -1
+            matchedUsers.push user
+            break
+    matchedUsers
+
 # Private: Extend obj with objects passed as additional args.
 #
 # Returns the original object with updated changes.


technicalpickles:
Interesting idea! I know I have done code to search multiple properties for a user, but they were specific properties, and I tried to prioritize specific ones.

Do you think that would be more useful? I'm imagining cases where some users have short usernames or whatever in some places, and multiple users would match, but only one is an exact match.

Would you mindg sharing the use case you prepared this for?

sdimkov:
The hubot that I'm making gathers information about the users in a chat service from various sources like spreadsheets and db-s for example. So the user objects contain many fields like personal email, mobile, extension, address, working floor.. etc. But they differ from user to user.

Given that not all users have the same fields, checking for specific fields and prioritizing would be very impractical in most of the cases.

So what I do instead is applying very simple logic:
1. users ask Hubot to do a task and provide details about users
  ex. Give deployment access to John Doe (matches 1 person with property name John Doe)
  ex. Give deployment access to P2 Team (matches 4 people with property team P2 Team)
2. Hubot searches for users with this universal search. If the search returns as many users as the task expects, Hubot does it. If the search returns more or less - Hubot replies to the sender to give better user details.

That eliminates the need of any defensive style programming and handling edge cases while still keeping very flexible search not just by name or email.

sdimkov:
@technicalpickles any movement on this?

I wish to submit some of the scripts from hubot as plugins but few of them depend on this.

technicalpickles:
Not yet, sorry. Traveling this week to give a talk on chatops at DevOpsDays
Rockies.

michaelansel:
cc #903, which adds a unit test foundation. If that gets merged before this, I'd like there to be unit tests for the new function.

As an aside, I feel like we should look at doing a refactor of the Brain to enable extensible user searching. Adding a new function to the core every time someone wants to process the data in a different way is...gross. No need to block this PR on that though.

sdimkov:
Guys can we now put some attention to this PR?

@michaelansel I agree that brain needs major redesign. But in the meantime this seems to me like the appropriate generic solution. And I would argue that it is definitely less gross than #usersForRawFuzzyName and #usersForFuzzyName :)

Just as in previous PRs I don't mind writing tests.

bhuga:
> Just as in previous PRs I don't mind writing tests.

This would be really helpful in moving this forward. The tests in #903 landed last week and should hopefully be a nice platform to build on.

src/brain.coffee
```diff
@@ -148,6 +148,21 @@ class Brain extends EventEmitter
...
+  usersForAny: (query) ->
+    query = query.toLowerCase()
+    matchedUsers = []
+    for key, user of (@data.users or {})
+      for property, value of user
+        if typeof value == 'string'
+          if value.toLowerCase().indexOf(query) isnt -1
+            matchedUsers.push user
+            break
+    return matchedUsers
```
technicalpickles: In :coffee:script, then `return` is implied.

src/brain.coffee
```diff
@@ -148,6 +148,21 @@ class Brain extends EventEmitter
...
+  # Public: Get all users that contain the provided query string
+  # in any of their properties.
+  #
+  # Returns an Array of User instances matching the search criteria.
+  usersForAny: (query) ->
+    query = query.toLowerCase()
+    matchedUsers = []
+    for key, user of (@data.users or {})
+      for property, value of user
+        if typeof value == 'string'
```
technicalpickles: You can use `is` for `==`, ie `if typeof value is 'string'`.

technicalpickles:
When thinking about new functionality, I am primarily considering it from the possibilities it adds, but also from the perspective of maintaining it long term, including ease of use and understanding for developers using it. My main concern about this is that it searches _all_ user attributes, and we don't know ahead of time what type of attributes adapters and scripts are going to stash in the brain for each user, so it could give confusing results without much insight into why. 

For example, let's say a script lets you add your favorite piece of furninture, and you add a script using this to search for people. If I said my favorite piece of furniture was a bench, and someone searches for 'Ben', my user is going to come up because `user.favorite_furniture` is `bench` for me, and it's completely unobvious to someone searching for "Ben" why "Josh Nichols" is coming up.

To your previous example, it sounds like there are a lot of different fields and users won't have them all, but there may be some well known ones (name, phone number, etc). To me, that says there should be a generic way to filter the brain by user, and possibly a more specific way to query a fixed set of attributes.

src/brain.coffee
sdimkov: I personally believe explicit returns are better than implicit ones except for single-line lambdas for example. Even for a terse ruby-influenced lang like coffeescript.

But given that the project uses implicit ones everywhere it makes sense to stick to the convention. I've fixed it.

src/brain.coffee
sdimkov: Fixed.

sdimkov:
@technicalpickles I understand the concern and the example provided is very good. I'm not really sure though if this illustrates a problem with the api itself or with incorrect usage of it.

Because if the end-user was surprised "Josh Nichols" showed up that highlights that the developer of the robot didn't document the people search well or didn't display enough details to the end-user. Also that the developer didn't think on the implications of using all-attributes search. In other words this is like saying regex is bad because somebody can write . without knowing it matches everything.
It is responsibility of the developer to understand the api-s he uses and to apply them correctly. Maybe adding a warning in the method description will help.

On the other hand your concern is still very valid. This method makes it easier to shoot yourself in the foot. Maybe that on its own is a strong enough argument to reject the PR. Maybe it is very easy to not think on the possible side effects of using this method.

I suggest you think on it all and make a call. Either let me know I should put some effort in tests, documenting risks, etc. Or choose to reject it. I'm ok with both decisions as I see reason in both.

technicalpickles:
@sdimkov thanks for the thoughtful response!

After mulling it over, I'd prefer not to add this given the concerns I've talked about already. It should be relatively easy to make helper methods on top of `robot.brain.users()` using `Array#filter` or underscore to pull out users matching what you want.

technicalpickles:
Oh, some random ideas I was considering too was an update to `robot.brain.users` to take a callback to filter the users, but again, I don't think that gives you much more than getting the users and filtering yourself.

michaelansel:
The only difference I see is around interfaces: if you are pulling from an internal data structure, that might break if we change the innards. If there is a fixed function interface separate from the data structure, we can test for and protect the interface separate from the data structure (c.f. the problems we are having with #917)

Issue #858, opened 2015-02-03
Allow programmatic launching of hubot


michaelansel:
As discussed in several other issues (#514, #768, #801, #815), there is a desire for a programmatic way of launching hubot to control script loading and to combine with existing node applications.

The proposed solution is something along the lines of [Rack](http://rack.github.io/)'s config.ru (which I'll reference as the "launch config file").

Primary design constraints:
- hubot should be able to be launched from an existing node.js application
- the launch config file should specify which scripts are to be loaded and in what order
- the launch config file should allow injecting arbitrary middleware/dependencies early in the process (e.g. load an external configuration file before any scripts that would use it)
- the launch config file should allow arbitrary JS/CS code for detailed configuration of complex extensions (e.g. load an authorization framework and then configure it inside the launch config file)
- the launch config file should have a mechanism for loading legacy scripts (assuming the script interface changes)

Not really in scope, but could we make it so that a robot can be attached to multiple adapters?

Terminology:
- `script` is the current style of scripts (`module.exports` is a `Function` that is called with the `robot` object)
- `module` is the new-style of scripts; details are undefined other than it can be more than just a `Function` and could have a namespace for all listeners it registers

An initial pass at the design:

``` coffee
module.exports = (robot) ->

# Adapter
robot.use require('hubot-irc')

# Middleware
robot.use require('hubot-ldap-user-info') # add extra fields to the msg.user object
robot.use require('hubot-rbac') # apply RBAC authorization rules

# Scripts/Modules
robot.use robot.wrapLegacyScript('mysql', require('hubot-mysql-chatops'))
robot.use require('hubot-duo') # accept two-factor keys and update global user info
```

Breakdown:

``` coffee
module.exports = (robot) ->
```

Maybe stick with the current script interface? This could make forward-compatibility possible (which would make moving to the rackup file optional and not require a major version bump).

``` coffee
# Adapter
robot.use require('hubot-irc')
```

Could we make adapters just another module? Not sure what the implications are for the underlying implementation.

``` coffee
# Middleware
robot.use require('hubot-ldap-user-info') # add extra fields to the msg.user object
robot.use require('hubot-rbac') # apply RBAC authorization rules
```

Load middleware in whatever order makes sense (since order matters for middleware)

``` coffee
robot.use robot.wrapLegacyScript('mysql', require('hubot-mysql-chatops'))
```

[...]

PR #924, opened 2015-04-20
Hooks

bhuga:
Hooks

This is a work-in-progress. I'm opening it early, because there's several parts I'd like feedback on.

This is inspired heavily by @michaelansel's work over in https://github.com/github/hubot/pull/803. We have been abusing `robot.respond /.*/` quite a bit in order to accomplish similar goals, and that pull was eye-opening. This pull does not include an equivalent of the listener metadata from that pull, because that pull does it so elegantly it does not seem worthwhile to revisit. We should get that into master no matter what.

The middleware itself, however, was a little uncomfortable for a few of our use cases. I played around with it and implemented a few things and came up with these issues:

 * They only apply to one phase of a message, "when a listener has matched." I found distinct use cases for at least 3 places (read on).
 * The name 'middleware' has a lot of context from its use in e.g. express, but these do a different thing. They don't have access to responses, for example, and an incoming message can create several responses. It's a hook, not a middleware.
 * It makes listeners async. This complicates testing and would be breaking the API enough that I think it would require calling it 3.0.
 * It added a dependency, and I always want to examine that closely.

This pull adds support for 3 hooks: `prereceive`, `prelisten`, and `prereply`. They run before any listeners have executed, before an individual listener executes, and before any message is sent, respectively.

An example of a `prereceive` hook that blacklists a room but for certain users:

```coffeescript
  robot.prereceive (hook) ->
    if blacklistedRoom(hook.message) && !whitelistedUser(hook.message)
      if hook.message.text.match(/^(\/|hubot)/) # if the message starts with / or 'hubot'
        hook.response.reply "I'm sorry, but I'm configured to ignore commands from this room."
      # But ignore everything, even hear commands.
      hook.finish()
```

An example of a `prelisten` hook, taken from #803. It assumes the listener middleware, which is not in this pull:

```coffeescript
 # node-statsd like interface
  stats =
    increment: (key) ->
      console.log "#{key}:1|c"
      # noop
robot.prelisten (hook) ->
    listenerId = hook.listener.options?.id
    if listenerId?
      stats.increment "hubot.listeners.total.#{listenerId}"
      if hook.response.message.user?.name?
        stats.increment "hubot.listeners.by_user.#{hook.response.message.user.name}.#{listenerId}"
      if hook.response.message.user?.room?
        stats.increment "hubot.listeners.by_room.#{hook.response.message.user.room}.#{listenerId}"
```

An example of a `prereply` hook. This prevents sending a message that matches `/passwords/`:
```coffeescript
robot.prereply (hook) ->
  if hook.reply.text.match(/passwords/)
    hook.finish()
```

And this changes the outgoing message:
```coffeescript
robot.prereply (hook) ->
  if hook.reply.text.match(/passwords/)
    hook.reply.text = "Sorry meatbag, no passwords."
```

I am pretty happy with the hook API as used, but I still need to do some more testing. I'm very interested in feedback, though.

I'm less happy with what happened in the guts. There are a few cases where things could be cleaned up internally if we refactored a few things. For example, `Response`s don't know about their `Listener`s, so a `prereceive` hook constructor needs a response, likewise a `prelistener` and its `Listener`. So the Hook API needs to handle both, but if responses had an when a smarter API might let it handle them a little more elegantly. If we like this direction, this pull would get a bit bigger as I did some reworking internally.

cc @technicalpickles

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -517,34 +517,6 @@ class Robot
     HttpClient.create(url, options)
       .header('User-Agent', "Hubot/#{@version}")
 
-  # Private. Run the hooks of a given name
-  #
-  #  name     - 'prereceive'
-  #  response - The response object
-  #  listener - The matching listener, if any
-  #  message  - A message that hubot is trying to send, if any
-  #
-  # Each hook must call hook.next() or hook.done(). Hook.next() continues
-  # processing and hook.done() aborts the message. If a response message
-  # exists, hook.done() prevents it from being sent.
-  #
-  # Returns nothing.
-  runHooks: (name, callback, message, response, listener, reply) ->
-    hooks = @hooks[name]
-    if hooks?.length > 0
-      opts =
-        hooks:    hooks
-        callback: callback
-        response: response
-        message:  message
-        listener: listener
-        robot:    this
-        reply:    reply
-      hook = new Hook(opts)
-      hook.run()
-    else
-      callback()
-
   # Public. Adds a prereceive callback hook to this robot.
   #
   #   cb  - A callback function which accepts a Hook object.
@@ -594,5 +566,33 @@ class Robot
     done = ->
       cb(response.envelope, string)
     @runHooks 'prereply', done, response.envelope.message, response, response.listener, string
+
+  # Private. Run the hooks of a given name
+  #
+  #  name     - 'prereceive'
+  #  response - The response object
+  #  listener - The matching listener, if any
+  #  message  - A message that hubot is trying to send, if any
+  #
+  # Each hook must call hook.next() or hook.done(). Hook.next() continues
+  # processing and hook.done() aborts the message. If a response message
+  # exists, hook.done() prevents it from being sent.
+  #
+  # Returns nothing.
+  runHooks: (name, callback, message, response, listener, reply) ->
+    hooks = @hooks[name]
+    if hooks?.length > 0
+      opts =
+        hooks:    hooks
+        callback: callback
+        response: response
+        message:  message
+        listener: listener
+        robot:    this
+        reply:    reply
+      hook = new Hook(opts)
+      hook.run()
+    else
+      callback()
 
 module.exports = Robot
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -61,8 +61,8 @@ class Robot
     @adapterName   = adapter
     @errorHandlers = []
 
-    @on 'error', (err, msg) =>
-      @invokeErrorHandlers(err, msg)
+    @on 'error', (err, res) =>
+      @invokeErrorHandlers(err, res)
     @onUncaughtException = (err) =>
       @emit 'error', err
     process.on 'uncaughtException', @onUncaughtException
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -529,13 +529,15 @@ class Robot
   # exists, hook.done() prevents it from being sent.
   #
   # Returns nothing.
-  runHooks: (name, callback, message, reply) ->
+  runHooks: (name, callback, message, response, listener, reply) ->
     hooks = @hooks[name]
     if hooks?.length > 0
       opts =
         hooks:    hooks
         callback: callback
+        response: response
         message:  message
+        listener: listener
         robot:    this
         reply:    reply
       hook = new Hook(opts)
@@ -552,5 +554,25 @@ class Robot
   prereceive: (cb) ->
     @hooks['prereceive'] ||= []
     @hooks['prereceive'].push cb
+
+  # Public. Adds a prelisten callback hook to this robot.
+  #
+  #   cb  - A callback function which accepts a Hook object.
+  #
+  # See also Hook
+  # Returns nothing
+  prelisten: (cb) ->
+    @hooks['prelisten'] ||= []
+    @hooks['prelisten'].push cb
+
+  # Protected. For use by Listener to run prelisten callbacks after a match
+  # is found.
+  #
+  # See also Hook
+  # Returns nothing
+  runPrelistenHooks: (listener, response) ->
+    done = ->
+      listener.callback(response) unless response.envelope.message.done
+    @runHooks 'prelisten', done, response.envelope.message, response, listener
 
 module.exports = Robot
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -575,4 +575,24 @@ class Robot
       listener.callback(response) unless response.envelope.message.done
     @runHooks 'prelisten', done, response.envelope.message, response, listener
 
+  # Public. Adds a prereply callback hook to this robot.
+  #
+  #   cb  - A callback function which accepts a Hook object.
+  #
+  # See also Hook
+  # Returns nothing
+  prereply: (cb) ->
+    @hooks['prereply'] ||= []
+    @hooks['prereply'].push cb
+
+  # Protected. For use by Response to run hooks before replying.
+  #
+  # TODO: we should have a listener available here, but there's no way to get
+  # to it. Responses belong to listeners but don't have a way to get to them,
+  # and we should change that API.
+  runPrereplyHooks: (response, string, cb) ->
+    done = ->
+      cb(response.envelope, string)
+    @runHooks 'prereply', done, response.envelope.message, response, response.listener, string
+
 module.exports = Robot
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -9,6 +9,7 @@ Brain = require './brain'
 Response = require './response'
 {Listener,TextListener} = require './listener'
 {EnterMessage,LeaveMessage,TopicMessage,CatchAllMessage} = require './message'
+{Hook} = require './hook'
 
 HUBOT_DEFAULT_ADAPTERS = [
   'campfire'
@@ -49,6 +50,7 @@ class Robot
     @listeners = []
     @logger    = new Log process.env.HUBOT_LOG_LEVEL or 'info'
     @pingIntervalId = null
+    @hooks     = {}
 
     @parseVersion()
     if httpd
@@ -191,17 +193,29 @@ class Robot
   #
   # Returns nothing.
   receive: (message) ->
+    curriedReceive = =>
+      @receiveWithoutHooks(message)
+    @runHooks 'prereceive', curriedReceive, message, null
+
+  # Private: receive a message, passing it to any inerested listeners. Does
+  # not execute callbacks.
+  #
+  # message - A Message instance
+  #
+  # Returns nothing.
+  receiveWithoutHooks: (message) ->
     results = []
     for listener in @listeners
       try
-        results.push listener.call(message)
         break if message.done
+        results.push listener.call(message)
       catch error
         @emit('error', error, new @Response(@, message, []))
 
         false
+
     if message not instanceof CatchAllMessage and results.indexOf(true) is -1
-      @receive new CatchAllMessage(message)
+      @receiveWithoutHooks new CatchAllMessage(message)
 
   # Public: Loads a file in path.
   #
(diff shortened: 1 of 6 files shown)


src/robot.coffee
```diff
@@ -393,26 +407,6 @@ class Robot
...
 
-  # Public: A helper send function which delegates to the adapter's send
-  # function.
-  #
-  # user    - A User instance.
-  # strings - One or more Strings for each message to send.
-  #
-  # Returns nothing.
-  send: (user, strings...) ->
-    @adapter.send user, strings...
```
bhuga: These methods violate the API contract of the adapters, sending a user instead of an envelope that wraps a user. I initially thought about using them instead of a response for `prereceive` middleware, but I ended up wanting to use a response so I could centralize handling `prereply` middleware.

They seem broken, though, so I went ahead and removed them. Tell me if I'm missing something?

src/robot.coffee
```diff
@@ -503,4 +497,80 @@ class Robot
...
+  #  response - The response object
+  #  listener - The matching listener, if any
+  #  message  - A message that hubot is trying to send, if any
+  #
+  # Each hook must call hook.next() or hook.done(). Hook.next() continues
+  # processing and hook.done() aborts the message. If a response message
+  # exists, hook.done() prevents it from being sent.
+  #
+  # Returns nothing.
+  runHooks: (name, callback, message, response, listener, reply) ->
```
bhuga: lol. I would like to clean this up by making response aware of messages and listeners, and letting `match` and `listener` potentially be null.

test/listener_test.coffee
```diff
@@ -17,6 +17,8 @@ describe 'Listener', ->
     @robot =
       # Why is this part of the Robot object??
       Response: Response
+      runPrelistenHooks: (listener, response) ->
+        listener.callback(response)
```
bhuga: We're stubbing out response, so I have worsened the problem for now. If we like the direction I would try to refactor the hook handling out of the listener and let the robot handle it all, and this could go away.

test/robot_test.coffee
```diff
@@ -225,3 +216,85 @@ describe 'Robot', ->
...
+
+    it 'runs a prelistener hook before each listener', (done) ->
+      testMessage = new TextMessage(@user, 'message123')
+      @robot.hear /^message123$/, (response) ->
+        expect(response.message.addedData).to.equal("added data")
+        done()
+      listener = @robot.listeners[0]
+      @robot.prelisten (hook) ->
+        hook.message.addedData = "added data"
+        expect(hook.listener).to.equal(listener)
```
bhuga: I initially had a version where you had to call `hook.next()` instead of optionally calling `hook.finish()`. This was cool because things were still synchronous, but you could make a message wait on an external API call, SMS 2-factor, or whatever else. It was pretty annoying in practice, though, because not calling either in a hook meant *nothing happened.* It was an easy gotcha from a development perspective, and with after scientific survey with zero respondents, I decided the external service case was not common enough to optimize for.

If we want that, though, we can do a better job by raising an error if a hook exits without calling either. Feedback wanted.

src/robot.coffee
technicalpickles: I don't think I've ever used this myself, but I did a google for 'hubot "robot.reply"' and see some references to it out in the wild. It might be better to deprecate it, if we want to avoid a major version bump.

bhuga: Hmm, the uses send an envelope instead of a user, so its just a docs issue. I need to consolidate routes to the adapter to insert hooks, though; so I will:

 * update docs
 * send response to here
 * move the logic I added to response here

michaelansel:
I'm super excited that you are thinking about using middleware!

TL;DR; I think your use cases can be supported with a combination of existing plans and existing modules (e.g. [hubot-proxy](https://github.com/Hammertime38/hubot-proxy)).

For some context, my long-term vision of the message processing chain with the ability for extension between each step:
- Message is received by the adapter and a new `Message` object created
  - hooks that permit tweaking the message (not thought out at all)
- Adapter hands the new message off to `Robot` for processing
  - receive middleware; executed once per `Message` object
- Find all matching `Listener`s and then for each:
  - execute listener middleware (once per matching `Listener`)
  - execute listener callback as the "final" middleware
  - execute the `done` chain in listener middleware (unwind the stack)
- execute the `done` chain in receive middleware (executed once per `Message`, after all `Listener` processing is complete)
- Notify the adapter that message processing is finished
  - Post-receive hooks in the adapter? (not thought out at all)

I've considered treating the entire chain as a series of middleware, but put that thought process off until we at least get listener middleware merged. For the most part, I would like hubot to become a series of components with a default configuration that you can rewire however you want (cc #858).

Now, addressing your specific comments on the shortcomings of middleware/#803:

> They only apply to one phase of a message, "when a listener has matched." I found distinct use cases for at least 3 places (read on).

Yup, that was intentional. #803 has already taken 6 months to get merged (and the idea has been in flight since last June through #724), so I don't want to make it any more complex than it already is. I've been waiting until #803 is merged before building receive middleware (though I think it is pretty straightforward after the last refactor of #803). I *think* all adapter interaction is covered by [hubot-proxy](https://github.com/Hammertime38/hubot-proxy), so that might be the correct place to put in synchronous pre/post-receive logic and output interception, but we would still need to refactor if async is desired.

> The name 'middleware' has a lot of context from its use in e.g. express, but these do a different thing. They don't have access to responses, for example, and an incoming message can create several responses. It's a hook, not a middleware.

Listener middleware functions as close to Express middleware as I could make it. You have access to all the context that is passed down to the listener and the ability to execute before and after the listener callback with shared scope. These extra features make listener middleware a strict superset of prelisten hooks, so if you would rather have a hooks interface, you can just write a piece of middleware to provide generic hook execution.

Middleware is a [standard design pattern](http://www.mariocasciaro.me/the-strange-world-of-node-js-design-patterns). I originally called the construct hooks (#724), but renamed it at @technicalpickles's request. In my mind, middleware also implies a fixed ordering, whereas hooks should probably be fairly order independent. I intend to use middleware for things like adding extra context to a message (e.g. LDAP groups), which needs to happen before the authorization middleware executes. I'm open to changing the name, but I do feel like middleware is the correct one.

> It makes listeners async. This complicates testing and would be breaking the API enough that I think it would require calling it 3.0.

I worked very hard to ensure that the public API changes are 100% backwards compatible. #803 should only require a minor version bump. The addition of async was intentional: JavaScript is an event-driven language and the complete inability to do any I/O in the message processing workflow is a problem for things like authorization policy that needs to talk to an LDAP server (#768). Finally, I would like to make a correction to your statement: listener processing may be async, but listener callbacks themselves are still synchronous. In the no-middleware or all-sync-middleware cases, all timing behavior should remain exactly the same (I think there is even a test case to protect that behavior...).

> It added a dependency, and I always want to examine that closely.

Added at @technicalpickles's request. If the added dependency is a problem, the middleware chaining is pretty trivial to replicate (I originally wrote it myself and then moved to using the async module).

michaelansel:
On a separate note, I'm really intrigued by the idea of having a Hook or Middleware object that gets passed in to the callback (instead of a collection of arguments). I want to noodle on that some more.

bhuga:
Thanks for the strong answer; this is the kind of feedback I was after. We're trying to accomplish the same things, so we can definitely get something done.

> I'm open to changing the name, but I do feel like middleware is the correct one.

You've got me convinced that your use of "middleware" is correct. In particular, the concept of strong ordering is really not hook-ish but it's very important to what we're trying to do.

> On a separate note, I'm really intrigued by the idea of having a Hook or Middleware object that gets passed in to the callback (instead of a collection of arguments). I want to noodle on that some more.

I landed on this because one of my goals was that middleware should work like other middleware. I should be able to insert middleware that blacklists a user before all of hubot or just a listener and have it work the same way; the information required does not change. Different information is available at different points in the pipeline, so the hook object lets that one middleware work in any situation where the information it needs is available instead of having to rewrite it to take slightly different arguments.

> hubot-proxy

Thanks for the tip here; I didn't know about it and they have a few other hooks that I would like to add. But:

 * We want something in core.
 * Situations handleable by the proxy should work like other middleware does. I don't want the docs to say "use middleware for these situations and hubot-proxy for other situations."
 * hubot-proxy's hooks are not chainable, i.e. there can be only one 'shouldSendMessage` hook.

> I worked very hard to ensure that the public API changes are 100% backwards compatible.

You did a good job of this, but I think the asynchronous listener dispatch is a breaking change. Calling `response.finish()` inside a listener callback no longer works the same way, and listeners no longer have a guaranteed order of execution. It's possible I'm misunderstanding.

> For some context, my long-term vision of the message processing chain with the ability for extension between each step:
	•	Message is received by the adapter and a new Message object created
	◦	hooks that permit tweaking the message (not thought out at all)
	•	Adapter hands the new message off to Robot for processing
	◦	receive middleware; executed once per Message object
	•	Find all matching Listeners and then for each:
	◦	execute listener middleware (once per matching Listener)
	◦	execute listener callback as the "final" middleware
	◦	execute the done chain in listener middleware (unwind the stack)
	•	execute the done chain in receive middleware (executed once per Message, after all Listenerprocessing is complete)
	•	Notify the adapter that message processing is finished
	◦	Post-receive hooks in the adapter? (not thought out at all)

Thanks for this! It helps a ton to see where you're going.

I have an honest question, one of which came up while I was writing this branch. Namely, what can a receive or listen middleware do at the `done` stage? Even if the listeners are dispatched synchronously, their callbacks can do things asynchronously, so the middleware can't dispatch a listener and act on the reply. These callbacks have the exact same information available as the pre-listen and pre-receive middlewares, respectively. I couldn't think of anything a post-listen middleware might do that a pre-listen middleware might not.

This weird asymmetry is what initially pushed me away from the word 'middleware'. I feel like this `done()` unwind step is how you landed on dispatching listeners asynchronously (is this correct?), but I'm not seeing where it's helpful to keep the stack around for unwinding. Definitely interested in your thoughts here.

michaelansel:
Glad to hear we are on the same page :grin: 

> the hook object lets that one middleware work in any situation where the information it needs is available instead of having to rewrite it to take slightly different arguments

I think this is what we get with the context object added in the latest refactor. The difference I see now is just that middleware has the next/done functions broken out instead of part of an object. Given the ability to override the done function, maybe it makes sense to leave it as a separate argument instead of attaching it to an object. I'm not really sure, but I'll keep thinking about it. Thoughts definitely welcome.

> - We want something in core.
> - Situations handleable by the proxy should work like other middleware does. I don't want the docs to say "use middleware for these situations and hubot-proxy for other situations."

100% agreement on ensuring the docs are simple and straightforward, and 80% agreement on putting this into the core. The remaining 20% is general uncertainty because I haven't dug into adapter interactions too much yet.

> - hubot-proxy's hooks are not chainable, i.e. there can be only one 'shouldSendMessage` hook.

Oh, how interesting. I didn't realize that :-( I'm game for figuring out how to roll most of the functionality of hubot-proxy into the core. Hopefully we can reuse the middleware implementation proposed in #803, but we can do something different if necessary.

> I think the asynchronous listener dispatch is a breaking change. Calling response.finish() inside a listener callback no longer works the same way, and listeners no longer have a guaranteed order of execution.

`message.finish` should be protected by [this test case](https://github.com/github/hubot/pull/803/files#diff-940ae8596db3e68bb8078aee945ba998R224). If you can help build a test case for bad behavior, I'm happy to fix it.

> what can a receive or listen middleware do at the done stage? Even if the listeners are dispatched synchronously, their callbacks can do things asynchronously, so the middleware can't dispatch a listener and act on the reply.

The only use case I've come up with off the top of my head is a cost-based rate limiter (if a command takes a long time to execute, wait longer before allowing re-execution). I think the primary value here over independent pre and post hooks is the ability to share scope between the pieces. You could also share information by attaching to the response object, but this definitely feels cleaner.

RE: the async bit, while the entire process has been moved to rely on callbacks, all ordering should remain the same. You just need to wait on the callback instead of waiting on the function to return.

I could also imagine a world in which we buffer all the content of a response and pass it through the post-listener middleware, but that would definitely be a breaking API change.

> I feel like this done() unwind step is how you landed on dispatching listeners asynchronously (is this correct?), but I'm not seeing where it's helpful to keep the stack around for unwinding.

It was definitely part of it. The biggest driver for me was the need to do an LDAP lookup in between the message arriving and executing the listener callback. Hopefully the above explained my thought process, but let me know if you are confused, and I'll try again.

bhuga:
(Sorry for the delay continuing this conversation; was at devopsdaysrox).

> message.finish should be protected by this test case. If you can help build a test case for bad behavior, I'm happy to fix it.

Okay, I've read the code in more detail and figured out how you did this. It's not an API change.

> I could also imagine a world in which we buffer all the content of a response and pass it through the post-listener middleware, but that would definitely be a breaking API change.

I tried to work this out and I didn't end somewhere I liked. Any reply can have the listener and thus message that created it, so there's little value in having it in a stack as opposed to a pre-reply. This is a big reason I came away from "middleware", though you still have me convinced with the ordering.

> It was definitely part of it. The biggest driver for me was the need to do an LDAP lookup in between the message arriving and executing the listener callback. Hopefully the above explained my thought process, but let me know if you are confused, and I'll try again.

Can I ask how you ended up using `nextTick`? An LDAP lookup should be asynchronous, so I'm not sure what it's doing?

bhuga:
@michaelansel any thoughts on the question about `nextTick`?

michaelansel:
Ah, shoot, sorry. This slipped off my radar due to DevOpsDays Austin.

The `nextTick` is there to keep the function API consistent (always asynchronous). Not strictly required, but recommended by the [node docs](http://nodejs.org/api/process.html#process_process_nexttick_callback). Original comment: https://github.com/github/hubot/pull/803#issuecomment-77296881

bhuga:
We had a long talk today, and #803 is closer to what we want than this, so I'm closing this out.

P R O G R E S S

PR #927, opened 2015-04-26
Fix ambiguous respond regex when robot has alias substring of full name.

sdimkov:
This pull request fixes a rare bug with respond regex.

When hubot is started with an alias that is a substring of the robot's name weird things happen. Messages addressing the robot via his name can get matched by alias and the remaining chars from the name are being appended to the pattern. That breaks some hubot scripts.

For example the plugin hubot-auth has the following respond regex:
`robot.respond /@?(.+) ha(s|ve) (["'\w: -_]+) role/i, (msg) ->`
So if we start hubot like this:
 `./bin/hubot --name william --alias will`
and try to give John a developer role
```
william> william john has developer role
Shell: iam john does not exist
william>
```
we will get surprised that Hubot can't find John because it looks for "iam John" where "iam" is a left-over from the robot name.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -100,8 +100,9 @@ class Robot
 
     if @alias
       alias = @alias.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&')
+      [a,b] = if name.length > alias.length then [name,alias] else [alias,name]
       newRegex = new RegExp(
-        "^\\s*[@]?(?:#{alias}[:,]?|#{name}[:,]?)\\s*(?:#{pattern})"
+        "^\\s*[@]?(?:#{a}[:,]?|#{b}[:,]?)\\s*(?:#{pattern})"
         modifiers
       )
     else


michaelansel:
We should refactor to pull the pattern building logic out of `respond` and then add some unit tests to ensure this doesn't break anything. The refactor was already done as part of #871 (specifically, https://github.com/michaelansel/hubot/commit/029ec478b14f0a008752f144c8f850a75fea7b35), so you may be able to just cherry-pick that commit over.

sdimkov:
I totally agree on the need to refactor the respond/hear/listen logic. https://github.com/michaelansel/hubot/commit/029ec478b14f0a008752f144c8f850a75fea7b35 applies to this PR without conflicts.

But I suggest we don't mix bug fixes and features/improvements. I would prefer to keep this PR as simple and small as possible and get it in master asap. The bug it fixes annoys me heavily.

sdimkov:
@technicalpickles can we merge this bug fix?

sdimkov:
Now that this project seems to wake up - can we have a look at this PR ?

It's still a valid bug and it continues to occur. 

I also don't mind providing a regression unit test that covers the problem and the fix.

technicalpickles:
> I also don't mind providing a regression unit test that covers the problem and the fix.

That would be helpful, thanks :sparkles:

sdimkov:
@technicalpickles I wrote unit tests for #respond in separate PR (#967) for easier testing / review.

Please review and hopefully merge both.

Tdowneydev (commented):
repository-open-graph-template
[gitlab-tree-master.zip](https://github.com/hubotio/hubot/files/8074185/gitlab-tree-master.zip)
[monalisaoctocat.github.io-master.zip](https://github.com/hubotio/hubot/files/8074191/monalisaoctocat.github.io-master.zip)

PR #930, opened 2015-04-29
Fix scripting link for real

jtallant:


technicalpickles:
:sob: https://github.com/github/hubot/issues/916

PR #935, opened 2015-05-07
Update scripting docs for crafting post requests

jcbantuelle:
When sending post data via JSON, not all servers will properly parse this by default. Explicitly providing the Content-Type header is more robust.


michaelansel:
:+1:

jcbantuelle:
@michaelansel Thanks for the approval! Is there anybody else I should be pinging as well to finalize this PR?

michaelansel:
@technicalpickles is the one to ping for merging

PR #936, opened 2015-05-07
Parse the commands for the external scripts

nounoursheureux:
With this modification the external scripts commands are parsed and added to the hubot-help output.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -261,9 +261,11 @@ class Robot
       if packages instanceof Array
         for pkg in packages
           require(pkg)(@)
+          @parseHelp(require.resolve(pkg))
       else
         for pkg, scripts of packages
           require(pkg)(@, scripts)
+          @parseHelp(require.resolve(pkg))
     catch err
       @logger.error "Error loading scripts from npm package - #{err.stack}"
       process.exit(1)


michaelansel:
I haven't had time to think about this (maybe this week?), but I'm very confused... I don't see problems with external scripts not being in the command list, and yet your change looks like it should be required. My guess (without investigation): the loader in every external module calls `parseHelp` for each script it loads.

technicalpickles:
I have a feeling this is one of those bits of functionality that has seen churn. I haven't found the old references, but I'm pretty sure this used to exist and was causing problems. In addition, we've encouraged script authors to include code in their `index.coffee` to parse help, so this would mean having duplicate help messages.

michaelansel:
cc #858 because help-parsing needs to be part of programmatic loading

technicalpickles:
If you look at the default hubot scripts, they all have a index.coffee like https://github.com/hubot-scripts/hubot-pugme/blob/master/index.coffee (makes sense since I used [generator-hubot](https://github.com/github/generator-hubot)). You'll notice they are calling `robot.loadFile`. [`robot.loadFile`](https://github.com/nounoursheureux/hubot/blob/externalscripts-parsehelp/src/robot.coffee#L212-L227) calls `@parseHelp`. 

That means as is, it'd result in duplicate help commands. I'm going to close because of that.

It does bring up some interesting potential ideas though. Using `@parseHelp(require.resolve(pkg))`, it means you could make a package script, and the script would just be the `main` as specified in `package.json`, meaning less boilerplate. I'd be :+1: to that idea, but the migration path would be tricky for all the script packages already out there.

michaelansel:
> Using @parseHelp(require.resolve(pkg)), it means you could make a package script, and the script would just be the main as specified in package.json, meaning less boilerplate.

Huh? You lost me...

technicalpickles:
>the script would just be the main as specified in package.json

referring to this main https://github.com/hubot-scripts/hubot-pugme/blob/master/package.json#L30

>  meaning less boilerplate.

If there was only a single script, and hubot handled parsing the docs of it, there wouldnt' be a need for https://github.com/hubot-scripts/hubot-pugme/blob/master/index.coffee

michaelansel:
Got it; thanks!

PR #938, opened 2015-05-09
Fix error message about deprecation in executing 'hubot -c'

ta1kt0me:
According to master's hubot/docs/index.md

> % mkdir myhubot
> % cd myhubot
> % yo hubot

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -73,6 +73,7 @@ if Options.create
   console.error "'hubot --create' is deprecated. Use the yeoman generator instead:"
   console.error "    npm install -g yo generator-hubot"
   console.error "    mkdir -p #{Options.path}"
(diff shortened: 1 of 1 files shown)


bin/hubot
```diff
@@ -73,6 +73,7 @@ if Options.create
   console.error "'hubot --create' is deprecated. Use the yeoman generator instead:"
   console.error "    npm install -g yo generator-hubot"
   console.error "    mkdir -p #{Options.path}"
+  console.error "    mv #{Options.path}"
```
technicalpickles: Should this be `cd` instead of `mv`?

ta1kt0me: Thanks, what you say.

michaelansel:
LGTM :+1:

PR #943, opened 2015-05-17
Updating local adapter docs links

adamgross42:


adamgross42:
I see this will break the rendering at https://hubot.github.com/docs/ :/

technicalpickles:

PR #868, opened 2015-02-17
Add robot.catchAllAddressed

michaelansel:
Closes #683

Adds a new listener type that responds to messages that were addressed to the robot and didn't match any other (non-catch-all) listeners. More details in the docs.

PR #949, opened 2015-05-22
Change instanceof to respect interfaces

fcsonline:
Hubot is a plugin/adaptor framework that hosts them as node modules.
Those modules have their own dependencies versions. By depending on
types rather than interfaces, you run into the exact issue caused by
peerDependency explained [here](https://github.com/npm/npm/issues/5080)

### DIFF
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -57,7 +57,7 @@ class TextListener extends Listener
   # callback - A Function that is triggered if the incoming message matches.
   constructor: (@robot, @regex, @options, @callback) ->
     @matcher = (message) =>
-      if message.type == 'text'
+      if message.type is 'text'
         message.match @regex
     super @robot, @matcher, @options, @callback
 
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -57,7 +57,7 @@ class TextListener extends Listener
   # callback - A Function that is triggered if the incoming message matches.
   constructor: (@robot, @regex, @options, @callback) ->
     @matcher = (message) =>
-      if message.match
+      if message.type == 'text'
         message.match @regex
     super @robot, @matcher, @options, @callback
 
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -57,7 +57,7 @@ class TextListener extends Listener
   # callback - A Function that is triggered if the incoming message matches.
   constructor: (@robot, @regex, @options, @callback) ->
     @matcher = (message) =>
-      if message instanceof TextMessage
+      if message.match
         message.match @regex
     super @robot, @matcher, @options, @callback
 
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -63,7 +63,7 @@ class TopicMessage extends TextMessage
   type: 'topic'
 
 class CatchAllMessage extends Message
-  type: 'all'
+  type: 'catchall'
 
   # Represents a message that no matchers matched.
   #
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -1,4 +1,6 @@
 class Message
+  type: 'message'
+
   # Represents an incoming message from the chat.
   #
   # user - A User instance that sent the message.
@@ -12,6 +14,8 @@ class Message
     @done = true
 
 class TextMessage extends Message
+  type: 'text'
+
   # Represents an incoming message from the chat.
   #
   # user - A User instance that sent the message.
@@ -27,7 +31,7 @@ class TextMessage extends Message
   # Returns a Match object or null.
   match: (regex) ->
     @text.match regex
-  
+
   # String representation of a TextMessage
   #
   # Returns the message text
@@ -40,6 +44,7 @@ class TextMessage extends Message
 # text - Always null.
 # id   - A String of the message ID.
 class EnterMessage extends Message
+  type: 'enter'
 
 # Represents an incoming user exit notification.
 #
@@ -47,6 +52,7 @@ class EnterMessage extends Message
 # text - Always null.
 # id   - A String of the message ID.
 class LeaveMessage extends Message
+  type: 'leave'
 
 # Represents an incoming topic change notification.
 #
@@ -54,8 +60,11 @@ class LeaveMessage extends Message
 # text - A String of the new topic
 # id   - A String of the message ID.
 class TopicMessage extends TextMessage
+  type: 'topic'
 
 class CatchAllMessage extends Message
+  type: 'all'
+
   # Represents a message that no matchers matched.
   #
   # message - The original message.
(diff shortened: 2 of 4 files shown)


fmvilas:
:+1:

enricostano:
:+1:

edulan:
:+1:

saimonmoore:
This makes so much sense. :100:

src/message.coffee
```diff
@@ -54,8 +60,11 @@ class LeaveMessage extends Message
 # text - A String of the new topic
 # id   - A String of the message ID.
 class TopicMessage extends TextMessage
+  type: 'topic'
 
 class CatchAllMessage extends Message
+  type: 'all'
```
michaelansel: This should probably be 'catchall' or something a little more specific than 'all'

src/listener.coffee
```diff
@@ -39,7 +39,7 @@ class TextListener extends Listener
   # callback - A Function that is triggered if the incoming message matches.
   constructor: (@robot, @regex, @callback) ->
     @matcher = (message) =>
-      if message instanceof TextMessage
+      if message.match
```
michaelansel: Nope, this won't work. Not all messages with a `match` method have a `@regex` property. You need to do `if message.type == 'text'`

michaelansel: Wait. That's a lie. I was thinking of `matcher`. I still think we should do the test for `type == 'text'` though. Much more explicit of the intent (we want the `TextMessage` contract, not the 'I have a match method' contract)

test/listener_test.coffee
```diff
@@ -117,21 +117,20 @@ describe 'Listener', ->
...
           testRegex = /test/
- 
+
           testListener = new TextListener(@robot, testRegex, callback)
           result = testListener.matcher(testMessage)
- 
+
           expect(result).to.not.be.ok
-          expect(testMessage.match).to.not.have.been.called
+          expect(callback).to.not.have.been.called
```
michaelansel: Why does this test case need to change? As a unit test, we should be mocking out all dependencies (i.e. `match`) and making sure they are not called inappropriately.

michaelansel:
Added comments on the code. I'm not sure I fully understand the need for this change though... The linked npm issue is pretty meaty; I'll try to read through it and wrap my brain around the change, but a summary of what is going on would be immensely helpful.

src/listener.coffee
fcsonline: It's just a duck typing, at the end is the same. I can change it if you are concerned about non text messages with `match` properties.

src/message.coffee
fcsonline: ack

test/listener_test.coffee
fcsonline: The test is wrong IMO. The test is mocking the match property for and `EnterMessage`. This is wrong. I can restore it if I change the TextMessage check in the previous comment.

fcsonline:
It's a really long story around dependencies and plugins like hubot does. If you add `instanceof` checks then you hit a big issue for the plugin development. I tried to explain the problem in the PR description.

Also is related with the *gotchas* section in  https://hubot.github.com/docs/adapters/development/

src/listener.coffee
michaelansel: I think my brain was broken yesterday. I no longer have strong feelings about this. Checking for a specific type makes the minimal behavior change though, so the current state is probably the right one.

test/listener_test.coffee
michaelansel: Ah, now I understand. The intent of the test was to ensure that `match` wasn't accidentally getting called on non-`TextMessage` objects. If we are duck typing, I agree that the stub is wrong. If not, then the test can safely ensure that unknown methods aren't called.

Now that I understand the intent, I think the test can stay this way or change back. No strong feelings.

michaelansel:
Code looks good, I still want to understand the problem before giving a :+1: though. I'll try to find some time to noodle on it next week after Velocity. Ping me if I haven't gotten back to this by the end of next week.

fcsonline:
Ok! No problem!

src/listener.coffee
```diff
@@ -39,7 +39,7 @@ class TextListener extends Listener
   # callback - A Function that is triggered if the incoming message matches.
   constructor: (@robot, @regex, @callback) ->
     @matcher = (message) =>
-      if message instanceof TextMessage
+      if message.type == 'text'
```
bkeepers: For consistency?, should this be `if message.type is 'text'`

technicalpickles:
I'd be interested to know how (if at all) this affects adapters that create their own subclasses of `Message`, like slack does:

* https://github.com/slackhq/hubot-slack/blob/master/src/message.coffee
* https://github.com/slackhq/hubot-slack/blob/master/src/listener.coffee

I've tried thinking through it a few times, and didn't spot any obvious problems, but I'd like to get it written out at some point.

technicalpickles:
@fcsonline there was a bit of work done last week on master so there's a conflict now. Would you mind merging in master and getting an update pushed here? Should be good to merge after that.

michaelansel:
Unless this is fixing a bug/preventing other work, I'd rather have a good understanding of why this is needed before it merges. If you understand it though, @technicalpickles, then go for it. We also might want to add documentation somewhere of the different message types and what they mean (so that new message classes have an idea of what type to use).

michaelansel:
Okay, did some more noodling and testing. I didn't realize that instanceof only looks at the immediate parent and not the full inheritance hierarchy:

```coffeescript
class Message
  @type = 'message'

class TextMessage
  @type = 'text'

msg = new Message()
tmsg = new TextMessage()

# tmsg instanceof TextMessage # true
# tmsg instanceof Message # false
```

This led me to a different question though: why are we ever testing message types? Is the only reason we care for Listener matching?

My answer: yes. So, therefore message types are a way to bucket Message objects for selection by Listeners.

This would result in the following mapping:

Message Type | Listener "Types"
-----------|-----------
text | hear, respond, [eavesdrop](https://github.com/github/hubot/pull/871)
enter | enter
leave | leave
topic | topic
catchAll | catchAll, [catchAllAddressed](https://github.com/github/hubot/pull/868)

Then, any time a new Message subclass is created, it needs to specify a type that will associate it with one of the above Listener types.

If we do this, are there then certain expectations around how a certain Message subclass behaves? If so, we need to call those out. As I see it right now, the only expectation is:

* a `text` message should have a `match` method that takes a regular expression

Finally, it looks like the TextListener class is the only thing imposing requirements on Message types.

I won't (yet) go so far as to make a recommendation (want to noodle more), but that's what I've got in my head so far.

fcsonline:
@michaelansel, for this reason I more confortable doing duck typing checks instead of instanceof for TextListener.

michaelansel:
> for this reason I more confortable doing duck typing checks instead of instanceof for TextListener.

I'm inclined to agree with you. As an aside, I'm thinking about nuking `TextListener` as a follow up to #986, but that needs some more thought.

I'm still trying to find time to really think hard about this, but I imagine it will be a little while before this gets to the top of my queue.

sergeyt:
:+1:

Issue #598, opened 2013-10-16
Unix Deployment does not include `init.d` Script Example


timfallmk:
I noticed in the moved to in-repo documentation, the Unix deployment guide was given a much "higher level" treatment. While "keep-alive" methods are mentioned, there is no  quick-start or example script like the `init.d` one given in the old wiki. Was this intentional?

technicalpickles:
We started a pattern of including example init scripts in the `examples` script, along with some information in docs/deploying/unix.md, in case anyone is interested in helping write those up.

PR #955, opened 2015-06-01
Update unix.md with systemd unit file example

praseodym:
Many Linux distros are moving to systemd, and since it's pretty easy to configure, I thought I'd be nice to add an example to docs.

### DIFF
--- a/docs/deploying/unix.md
+++ b/docs/deploying/unix.md
@@ -53,11 +53,13 @@ really, you probably want some process monitoring using tools like
 [monit](http://mmonit.com/monit/),
 [god](http://godrb.com/),
 [bluepill](https://github.com/arya/bluepill),
-[upstart](http://upstart.ubuntu.com/).
+[upstart](http://upstart.ubuntu.com/),
+[systemd](http://freedesktop.org/wiki/Software/systemd/).
 
 For starting at boot, you can create an init script for appropriate your UNIX
 distribution, or if you are using one the process monitoring tools above, make
-sure it boots at startup.
+sure it boots at startup. See the [examples](https://github.com/github/hubot/tree/master/examples)
+for configuration examples.
 
 ## Recommendations
 
--- a/examples/hubot.service
+++ b/examples/hubot.service
@@ -1,0 +1,23 @@
+; Hubot systemd service unit file
+; Place in e.g. `/etc/systemd/system/hubot.service`, then `systemctl daemon-reload` and `service hubot start`.
+
+[Unit]
+Description=Hubot
+Requires=network.target
+After=network.target
+
+[Service]
+Type=simple
+WorkingDirectory=/path/to/hubot
+User=change-to-hubot-user
+
+Restart=always
+RestartSec=10
+
+; Configure Hubot environment variables, make sure to use quotes around whitespace as shown below.
+Environment=HUBOT_aaa=xxx "HUBOT_bbb='yyy yyy'"
+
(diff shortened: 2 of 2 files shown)


docs/deploying/unix.md
```diff
@@ -53,12 +53,39 @@ really, you probably want some process monitoring using tools like
...
+Restart=always
+RestartSec=10
+
+Environment=HUBOT_aaa=xxx HUBOT_bbb=yyy
+
+ExecStart=/opt/hubot/bin/hubot --adapter zzz
+
+[Install]
+WantedBy=multi-user.target
+```
```
bkeepers: There currently aren't example scripts for the other tools. Does it make sense to include on only for systemd?

praseodym: I would certainly welcome init scripts for other tools as well. Systemd has the advantage of being the default init system of most Linux distributions, though.

technicalpickles:
cc https://github.com/github/hubot/issues/598 where there's talk of documenting an init.d script

technicalpickles:
I think the hard part with having specific examples is that it's really hard to do well for this kind of thing.

If you are too specific (ie include stuff like paths for /opt/hubot), then it's not necessarily work for everyone. If you are too general, then it's easy to overlook small details. Either way, that means more people have problems getting hubot running, which means more support requests at best, or less hubot deployment at worst.

We currently only use god for managing the process, so it's hard to give good feedback on this, and also it'd be hard to keep this up to date.

praseodym:
Since these are only examples, not files to be used verbatim, I think it should be fine to include paths.

Users can still deploy using whatever deployment method they like, I just want to make it a bit easier to deploy properly instead of e.g. using screen manually.

technicalpickles:
One thought I had was to create an `examples` directory for hubot, and include things like example scripts and other configuration. The systemd script might be a good one to start with. Once that's there, this documentation can be trimmed a bit, and just link to the example. The example could also be annotated with code comments to indicate any points that need to be adjusted.

@praseodym does that seem reasonable?

praseodym:
Sounds good! :shipit:

technicalpickles:
@praseodym this was your PR, so will you be making the change I suggested?

praseodym:
Check it out :)

docs/deploying/unix.md
```diff
@@ -53,11 +53,12 @@ really, you probably want some process monitoring using tools like
...
 [god](http://godrb.com/),
 [bluepill](https://github.com/arya/bluepill),
-[upstart](http://upstart.ubuntu.com/).
+[upstart](http://upstart.ubuntu.com/),
+[systemd](http://freedesktop.org/wiki/Software/systemd/).
 
 For starting at boot, you can create an init script for appropriate your UNIX
 distribution, or if you are using one the process monitoring tools above, make
-sure it boots at startup.
+sure it boots at startup. See the `examples` for configuration examples.
```
michaelansel: How about making this a link to the folder?

```
See the [examples](https://github.com/github/hubot/tree/master/examples) for configuration examples.
```

See the [examples](https://github.com/github/hubot/tree/master/examples) for configuration examples.

praseodym: Done!

michaelansel:
:+1: I think this addresses all the comments. What do you think, @technicalpickles?

technicalpickles:
Looks good. Thanks @praseodym !

PR #967, opened 2015-06-06
Implement #respond unit tests. Covers bug described in PR #927

sdimkov:
Covers functionality of robot.respond. Provides regression check for bug described in #927

Please note that 1 of the 5 new tests fails in the current master. That is because there is bug that it highlights. Once #927 is applied all tests start passing again.

@technicalpickles @michaelansel


michaelansel:
These are all integration tests for the `Robot` class, not unit tests. I'll come back and reorganize them later. Thanks for thinking through the flows though!

PR #975, opened 2015-06-10
Refactor respond for testability

michaelansel:
Pull the pattern building logic out so we can focus on testing just the resulting regex

This converts the tests added in #967 over to proper unit tests.

@sdimkov Can you eyeball these tests and confirm they still protect against regressions in #927? They appear to work correctly when I revert the merge, but extra :eyes: are always good.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -91,6 +91,15 @@ class Robot
   #
   # Returns nothing.
   respond: (regex, options, callback) ->
+    @listeners.push new TextListener(@, @respondPattern(regex), options, callback)
+
+  # Private: Build a regular expression that matches messages addressed
+  # directly to the robot
+  #
+  # regex - A RegExp for the message part that follows the robot's name/alias
+  #
+  # Returns RegExp.
+  respondPattern: (regex) ->
     re = regex.toString().split('/')
     re.shift()
     modifiers = re.pop()
@@ -116,7 +125,7 @@ class Robot
         modifiers
       )
 
-    @listeners.push new TextListener(@, newRegex, options, callback)
+    return newRegex
 
   # Public: Adds a Listener that triggers when anyone enters the room.
   #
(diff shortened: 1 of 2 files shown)


michaelansel:
The diff looks gross, but I basically just rewrote the tests from #967 as unit tests instead of integration tests.

sdimkov:
:+1: This makes total sense.

I did review the code and also tested with/without the fix. All good.

src/robot.coffee
```diff
@@ -116,7 +125,7 @@ class Robot
         modifiers
       )
 
-    @listeners.push new TextListener(@, newRegex, options, callback)
+    return newRegex
```
sdimkov: According @technicalpickles 's feedback on https://github.com/github/hubot/pull/922 the return statements should be implicit.

michaelansel: You are absolutely right. Fixed

michaelansel:
@technicalpickles Can you give this some :eyes: since I'm adding a new private method?

technicalpickles:
@michaelansel looks good to me :+1:

sdimkov:
@michaelansel  what do you think on marking `robot#respondPattern` as a public method ?

I have use-case where my adapter needs to know if a message is directly pointed to the robot or not. So I currently call `robot#respondPattern` which is broken in the long term as this being private gives me no guaranty that the method won't get changed or removed in future.

I believe it is best to just make it public. But if you think otherwise we can add another simpler method for public access. Something like `robot#shouldRespond(message) ->` which returns _boolean_

sdimkov:
cc #1002

technicalpickles:
>  what do you think on marking `robot#respondPattern` as a public method ?

I could see that. I was trying to make arguments for keeping it private, and the only reason I could see if we don't like some detail of it or we expect it to change heavily. I don't see that for `respondPattern` though.

Issue #965, opened 2015-06-05
Refine and automate release process


technicalpickles:
I walked through the process with @bhuga, @bkeepers, and @michaelansel and there is definitely some tedious points.

See  https://github.com/github/hubot/blob/master/CONTRIBUTING.md#releasing for the current process. Particularly tedious things:
- [ ] generating ChangeLog summary from PRs merged
- [ ] generating a PR with links to PRs, cc author, etc
- [ ] push a prerelease package (`hubot@next`) and execute functional tests
- [ ] promote prelease package (`npm dist-tag add hubot@<version> latest`)
- [ ] update hubot.github.com's submodule of github/hubot docs directory

jasonkarns:
marginally related: any particular reason why there is a release script when npm has `npm version` built in?

jasonkarns:
I guess the general question is why there are so many bash scripts in the first place? Most all of them could be replaced (and simplified) with npm scripts. For instance, nearly every bash script has to mess with modifying the PATH so that node_modules bins are available. That is unnecessary when using npm scripts since npm prepends `$(npm bin)` to the path automatically.

This is already a node project, and it would be more conventional to have `npm test`, `npm start` and see all the other available scripts with `npm run`.

I would be happy to put together PR illustrating the change, if it would be considered.

technicalpickles:
> any particular reason why there is a release script when npm has npm version built in?

Yes, the reason is we didn't know about it :grin: 

For what it's worth, we are currently updating package.json separate from actually pushing it. By doing this, we can do a PR which includes the PR, and then publish it after.

technicalpickles:
>  I guess the general question is why there are so many bash scripts in the first place? Most all of them could be replaced (and simplified) with npm scripts. For instance, nearly every bash script has to mess with modifying the PATH so that node_modules bins are available. That is unnecessary when using npm scripts since npm prepends $(npm bin) to the path automatically.

Part of it is conventions used at GitHub for development. Nearly every project has a `script/bootstrap` for initializing the local environment, `script/test` for running test, `script/server` for launching the server process.

Another part is that hubot is old enough that I don't think these conventions were as well established when it was written, and they just haven't been updated.

> This is already a node project, and it would be more conventional to have npm test, npm start and see all the other available scripts with npm run.

I'd be amenable to using these conventions. It would be good to keep some of the scripts around though for consistency with our other projects.

---

[...]

PR #978, opened 2015-06-11
Utilizing npm-scripts instead of make and scripts/*

jasonkarns:
This isn't really a merge-ready PR but mostly to illustrate what straight npm-scripts can do with much less configuration. I also find it more discoverable than a directory of shell scripts or make targets because an npm project _always_ has a package.json, and therefore npm-scripts are there by default. 
- scripts/bootstrap still remains, though it is largely unnecessary (all it does now is source .env files, which probably ought to be it's own npm utility and thus easily done directly in npm-scripts)
- bootstrap is run automatically before `test` and `start` scripts (via `pretest` and `prestart`)
- `start` is the old script/server
- `test` is the mocha test suite (script/test and make test)
- `clean` removes the lib directory (coffee-compiled JS); was the old make remove-js target
- `test:smoke` runs the smoke-test (was script/smoke-test)
- `release` is the old script/release. It's broken into three steps: prerelease (which does the npm-update and smoke-test); then 'release' via npm-version (bump/tag/commit); and finally postrelease (npm-publish and git-push). If any step fails, the subsequent steps are aborted.
- `compile` is the old generate-js make target
- `compile:watch` is the old make dev target (could easily be renamed but I think this is clearer)
- `package` is the old make package target

Other notes:
- npm-scripts provide tab completion (`npm run com<TAB>` -> `npm run compile`)
- `npm run` prints a list of all configured scripts for easy usage/help, showing what's available
- npm-scripts automatically prepend `node_modules/.bin` to `$PATH` so PATH manipulation is unnecessary
- npm-scripts support package.json introspection via environment variables that map directly to the package.json structure (e.g. `$npm_package_version`)
- npm-scripts support user/custom configuration with global and local `.npmrc` files that are quite useful. (i.e. configuring test reporters, for instance. so for example, mocha default reporter could be `spec`, but overridden on CI to `tap` while Rando Dev could configure theirs to be `dot`)


jasonkarns:
discussion spawned from #965

bkeepers:
:cool: I like this direction.

Although, may I make one request: can we keep the `script/*` files, even if they only call out to `npm foo`? We have a pretty strong convention in all of our projects at GitHub that you can just clone them and run `script/bootstrap; script/server` and be up and running. This convention makes it easy for anyone to play around with a project regardless of technology.

michaelansel:
I'm in favor of most of this; my primary concern is around the release process. There are a few sanity checks built into `script/release` to protect the blind (us) from walking off a cliff. None of the maintainers are heavy node.js developers, so extra checking is really nice. Also, we are planning to automate beyond just the simple mechanics (generate the changelog and PR for the release), so we need places to put in all that logic. Overall, I think we are going to need something like `script/release` in the long run.

jasonkarns:
@bkeepers which scripts are expected to exist?

- [ ] cibuild
- [ ] release
- [ ] server
- [ ] smoke-test
- [ ] test

I'm assuming that `cibuild`, `release`, and `test` are the primary ones?

bkeepers:
`bootstrap`, `scerver` and `test` are the primary ones. `cibuild` is only if there is special config for the CI environment.

jasonkarns:
recreated wrapper scripts `test` and `server` which call through to `npm test` and `npm start`, respectively.

the old `cibuild` script wasn't doing anything except calling down to `script/test` so I don't think `cibuild` is necessary.

release will need some additional work.

I am curious on the inclusion of `npm install` in the bootstrap script. That seems to slow things down quite a bit and is a bit unnecessary, IMO. Is `script/bootstrap` generally run manually? It's being invoked prior to both npm-start and npm-test.

If bootstrap is not generally run manually, then I would think the npm-install command could be removed from the bootstrap (which leaves its only duty to be sourcing `.env` files and setting `NODE_ENV`). That would speed up npm-test and npm-start (and by extension, script/server and script/test).

Alternatively, we could move npm-install out of bootstrap and into just the script/server and script/test scripts. Which would presumably be used by the non-node friendly devs (for which having npm-install run automatically is beneficial). For any node peeps, npm-install is likely almost muscle memory. So moving it into the `script/` scripts would give a speed boost to those who execute the npm scripts directly (via npm start/test).

It's mostly splitting hairs here, and it probably doesn't matter much either way. I just think a node person would find it especially odd that `npm test` and `npm start` trigger an `npm install` as part of their run.

bkeepers:
> I am curious on the inclusion of `npm install` in the bootstrap script. That seems to slow things down quite a bit and is a bit unnecessary, IMO. Is `script/bootstrap` generally run manually?

Yeah, I think it's fine to expect people to run `script/bootstrap` manually (as long as it's documented). We do have some apps that will run it automatically if needed, but they usually do some workarounds to check if it's really needed.

> Alternatively, we could move npm-install out of bootstrap and into just the script/server and script/test scripts. Which would presumably be used by the non-node friendly devs (for which having npm-install run automatically is beneficial). For any node peeps, npm-install is likely almost muscle memory. So moving it into the script/ scripts would give a speed boost to those who execute the npm scripts directly (via npm start/test).

That sounds reasonable to me as well.

jasonkarns:
Extracted .env sourcing into script/dotenv which is configured in npm to run via the preinstall, prestart and pretest hooks. Which means `.env`s will be sourced whether the tests are run via `npm test` or `script/test`. The same is true for `npm install` vs `script/bootstrap` and `npm start` vs `script/server`.

`script/bootstrap`, now that the .env stuff is moved into `script/dotenv`, is nothing more than a call to `npm install`. And it must be done manually. `script/bootstrap` is no longer executed automatically before a test run or server start.

jasonkarns:
This PR wasn't originally meant to be a mergeable thing, but rather to spark discussion on how npm scripts could be used to reduce some script code and leverage the built in functionality. With the exception of `release`, though, this thing is actually starting to look shippable. Maybe we should yank the `release` changes out of this PR and make that it's own (smaller, reviewable) thing?

technicalpickles:
> cibuild is only if there is special config for the CI environment.

We're using Travis CI for this project now, so it's not needed anymore.

technicalpickles:
> Maybe we should yank the release changes out of this PR and make that it's own (smaller, reviewable) thing?

That is generally a good plan :+1:

technicalpickles:
> Alternatively, we could move npm-install out of bootstrap and into just the script/server and script/test scripts. Which would presumably be used by the non-node friendly devs (for which having npm-install run automatically is beneficial). For any node peeps, npm-install is likely almost muscle memory. So moving it into the script/ scripts would give a speed boost to those who execute the npm scripts directly (via npm start/test).

Just for some context, we use this script pattern for our projects across different languages. Running `npm install` might be muscle memory for a node developer, but it's not for a ruby developer (`bundle install`). In addition, I think you'll find developers often add local shell aliases or commands for running `npm install` before `npm start`, which doesn't help anyone else but themself. So that is why we use them: for consistency across projects & languages, and to give everyone the same tools for interacting with the projects.

jasonkarns:
Okay, this stalled for a bit... but I've reverted the changes to `script/release`.

So now the scripts are:

`script/bootstrap` delegates to `npm install`
`script/server` delegates to `npm start`
`script/test` delegates to `npm test`
`script/smoke-test` delegates to `npm run test:smoke` (not sure if the smoke-test script is even run or if we can kill it entirely)

the additional npm scripts include:
`npm run clean`

- removed makefile
- removed script/cibuild

Lastly, the .env sourcing behavior that *used* to live in `script/bootstrap` is now in `script/dotenv` which is invoked `preinstall`, `prestart` and `pretest`

The release stuff will be saved for a separate PR. This one should be reviewable to merge (though likely needs rebased now) :/

jasonkarns:
rebased onto latest master

michaelansel:
Overall, I'm a :+1: on this. Only outstanding question: I see you are adding a new `package` command and using it to generate a new hubot, which adds dependencies on `yo` and `generator-hubot`. Why? Is this intended to be a smoke test or what? As a core developer, that's not a task I frequently perform...

jasonkarns:
@michaelansel the `package` script came from the Makefile. https://github.com/github/hubot/blob/54c6be9a8857521964024a65a7d0358da82f3de6/Makefile#L14-L16 (as did the `clean` script)

If that's not needed, I can definitely remove it and the deps.

package.json
```diff
@@ -26,11 +26,13 @@
     "scoped-http-client": "0.11.0"
   },
   "devDependencies": {
+    "generator-hubot": "^0.3.1",
```
bkeepers: This will bring in way too may dependencies. Let's remove this and whatever functionality depends on this.

package.json
```diff
@@ -26,11 +26,13 @@
...
     "hubot-mock-adapter": "^1.0.0",
     "coffee-errors": "0.8.6",
     "mocha": "^2.1.0",
     "mockery": "^1.4.0",
-    "sinon-chai": "^2.6.0"
+    "sinon-chai": "^2.6.0",
+    "yo": "^1.4.7"
```
bkeepers: Same here.

Issue #977, opened 2015-06-10
hubot package installs with `.node-version` breaking nodenv


jasonkarns:
The npm package is published with a `.node-version` file. I can't seem to find when, how or where this file is generated, but I humbly submit that it should not be included in the package (nor the git repo).

As has been discussed in the ruby community numerous times, the .node-version file should not be shared as it requires an overly strict version specifier.

Does anyone know where this file is coming from before I go further in advocating for its removal?

technicalpickles:
I don't see it in git:

```
$ git rm .node-version
fatal: pathspec '.node-version' did not match any files
```

I generated a new bot and checked it's using 2.13.1. I did the release over the weekend, and I don't have that file locally, so I'm not sure where it's coming from.

Can you demonstrate how this is breaking nodenv? I use it myself, and have not seen a problem, since it'd be inside a projects's `node_modules`, not it's parent.

jasonkarns:
I'm terribly confused where it's coming from:

```
$ npm i hubot
$ ls -al node_modules/hubot/.node-version 
-rw-r--r--  1 jasonkarns  wheel     6B Nov 21  2014 node_modules/hubot/.node-version
```

(modified time of Nov 21, 2014? weird)

I verified it's in the actual package from npm: http://registry.npmjs.org/hubot/-/hubot-2.13.1.tgz

It breaks on my nodenv because the contents of the node-version aren't a valid version name (`v0.10`). nodenv version specifiers don't have the `v` prefix, and include the full major.minor.patch version number.

```
$ bin/hubot 
/Users/jasonkarns/Projects/td/doubot/node_modules/hubot/node_modules/.bin
nodenv: version `v0.10' is not installed
```

technicalpickles:
Did you generate a hubot with `yo hubot`? Try running `bin/hubot` from `/Users/jasonkarns/Projects/td/doubot` instead.

technicalpickles:
Also, turns out there are multiple nodenvs: https://github.com/wfarr/nodenv

jasonkarns:
> Try running bin/hubot from /Users/jasonkarns/Projects/td/doubot instead.

That *is* where I'm running `bin/hubot` from:

```
$ pwd
/Users/jasonkarns/Projects/td/doubot
$ bin/hubot 
/Users/jasonkarns/Projects/td/doubot/node_modules/hubot/node_modules/.bin
nodenv: version `v0.10' is not installed
```

[...]

PR #980, opened 2015-06-12
git-ignore .node-version

jasonkarns:
There is no .npmignore so npm will fallback to .gitignore to determine
which files to omit from the package.


jasonkarns:
fixes #977

PR #982, opened 2015-06-15
Link to NPM docs on installing node/npm

### LINKED ISSUE #979: Trying to follow guide to getting started with Hubot, results differ from what the guide says.
So i am trying to follow the guide to simply install Hubot on my Ubuntu (14.04) server, but things do not turn out the way the guide says it will do. I've set up my server according to this guide: https://www.linode.com/docs/getting-started

And it looks like everything is ready get started.

Steps.

1. The guide says to follow another guide which says i need a compiler, i guess i can just use "sudo aptitude install build-essential" and i'm good to go?

2. "which gcc" now reveals that i now do have a compiler! (/usr/bin/gcc)

3. I use the first gist to install NPM and node.js and i then have to wait a really long time for all the text in my terminal to settle.

4. There is a line in the gist that did not execute: "curl https://www.npmjs.org/install.sh | sh", and when i try to execute it, i get this message:  
 % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                                       Dload  Upload   Total   Spent    Left  Speed
100    51  100    51    0     0    291      0 --:--:-- --:--:-- --:--:--   293
sh: 1: Moved: not found

I don't really know what it means, but at least i can try to just run the commands in install.sh manually and see what happens, right?

5.So i tried running it, and i got alot of text appearing. Okay, something happened. Woo! But then i get this message:

npm cannot be installed without node.js.
Install node first, and then try again.

Maybe node is installed, but not in the PATH?
Note that running as sudo can change envs.

PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

So it looks like install.sh does not know where my node-files are. Do you have any tips on what i can do further? The guide you have here does not really help that much for new users with little experience.

Aaaaand it looks like something happened to my ssh, making me unable to ssh into my server.

Am i just being to much of a noob to set up Hubot properly or is something in the guides you provide not functioning as intended?

bkeepers:
Closes #979

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,10 +6,7 @@ layout: docs
 
 ## Getting Started With Hubot
 
-You will need [node.js](http://nodejs.org/) and [npm](https://npmjs.org/). Joyent has
-an [excellent blog post on how to get those installed](http://joyent.com/blog/installing-node-and-npm), so we'll omit those details here.
-
-Once node and npm are ready, we can install the hubot generator:
+You will need [node.js and npm](https://docs.npmjs.com/getting-started/installing-node). Once those are installed, we can install the hubot generator:
 
     %  npm install -g yo generator-hubot
 


bkeepers:
Installing node is much easier than it was when the old docs were written. It also seems better to link to docs on node/npm's site, which should be kept up to date.

michaelansel:
:+1:

Issue #950, opened 2015-05-22
robot.respond to respond only once


ambikads:
I am new to hubot. I am building a chat program where the hubot asks series of questions and user replies with answers. The regex used could be the same in hubot's questions. The user reply should only match the asked question and not all the matches. How can I accomplish this? 

example
Hubot> Your first name?
User> anything
Hubot> Your last name?
User> Anything

The regex I use is just (.*). This matches both the questions here and I get replies from both the listeners.

Thanks.

michaelansel:
The short answer is: it's hard and there isn't a pre-existing pattern for this type of behavior.

If you're interested in putting in some development work, I imagine there is an opportunity to make a pretty cool framework to enable Q&A scripts. With the addition of listener options (#801, #803), you could register a listener, wait for a single response, and then destroy the listener. You might also need complex matchers (side effect of #871) so that another user's message or a message in another room isn't mistakenly matched.

No clue when the linked PRs will merge, but you can probably start developing against the general construct and tweak the interface if it changes.

michaelansel:
#801 has merged and I pulled the complex matcher logic out of #871 into #986. As soon as #986 merges, you should be good to go for building this.

michaelansel:
cc #319 which was also looking at ways to restrict listeners to specific rooms (or other complex conditions)

mose:
Seems `robot.listen` brought a nice way out of this issue. I'm closing it.

technicalpickles:
I've seen https://github.com/lmarkus/hubot-conversation come up a few times which is in this space.

I could see something like this becoming a more first class part of hubot. If anyone is interested in working on that, [Hubot Evolution](https://github.com/hubotio/evolution) is how we are handling new feature proposals.

michaelansel:
Cross referencing a SO answer I just put together on the same topic: https://stackoverflow.com/questions/43600466/hubot-nesting-commands/44145694#44145694

Issue #813, opened 2014-10-30
Creating a dynamic list of robot.respond at runtime?


victorhooi:
_NB: I'm not sure if this is quite in-scope for this issue tracker, please let me know if it's not._

Is there an easy way within Hubot to create a list of dynamic `robot.responds` at runtime?

For example, I'd like to retrieve a set of keys from a remote HTTP API (JIRA project codes, in this case).

I'd then like to create a `robot.respond` for each of the returned keys. Is there a recommended or idiomatic way of doing this in JS/Hubot?

michaelansel:
Reopening and self-assigning. I want to redirect this towards "dynamic matchers" -- I should be able to have a combination of regex and code to determine if a Listener should match. This give the appearance of dynamic listener registration. Kind of related to #871 (could leverage the refactor to generalize listener registration).

michaelansel:
This could be enabled by building on top of #986. There would likely be a little code duplication around regex matching, but I don't think that is terrible. All of the complicated logic (generating a respond pattern) is readily available.

Alternatively, you could use the idea @technicalpickles originally proposed now that #801 has merged.

bkeepers:
> This could be enabled by building on top of #986

Closing since this can be built on top of existing functionality.

Issue #319, opened 2012-07-20
Respond/Hear Conditions


tombell:
Maybe it is a good idea to add the option of specifying additional conditions to listeners in scripts.
- Specific rooms
- Specific users 

``` coffeescript
robot.respond /foo/, { user: ['tomb', 'atmos'], room: 40201 }, (msg) ->
  # ...
```

I am experimenting with the syntax due to maybe switching the way scripts are defined. See #249.

PR #986, opened 2015-06-18
Add Robot.listen

michaelansel:
I got annoyed with #871 and decided to rip out the refactor by itself.

Allow the creation of Listeners with custom match logic
Refactor to consolidate array manipulation logic

Known issue: Robot.hear still creates TextListeners. Eventually, I'd like to get that moved over to just a standard Listener with special match logic, but need to decide what to do with the potential interface changes.

**This introduces `Robot.listen` as a new public method**

### DIFF
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -77,3 +77,34 @@ module.export = (robot) ->
   robot.globalHttpOptions.httpAgent  = proxy('http://my-proxy-server.internal', false)
   robot.globalHttpOptions.httpsAgent = proxy('http://my-proxy-server.internal', true)
 ```
+
+## Dynamic matching of messages
+
+In some situations, you want to dynamically match different messages (e.g. factoids, JIRA projects). Rather than defining an overly broad regular expression that always matches, you can tell Hubot to only match when certain conditions are met.
+
+In a simple robot, this isn't much different from just putting the conditions in the Listener callback, but it makes a big difference when you are dealing with middleware: with the basic model, middleware will be executed for every match of the generic regex. With the dynamic matching model, middleware will only be executed when the dynamic conditions are matched.
+
+For example, the factoid lookup command could be reimplemented as:
+
+```coffeescript
+module.exports = (robot) ->
+  # Dynamically populated list of factoids
+  facts =
+    fact1: 'stuff'
+    fact2: 'other stuff'
+
+  robot.listen(
+    # Matcher
+    (message) ->
+      match = message.match(/^~(.*)$/)
+      # Only match if there is a matching factoid
+      if match and match[1] in facts
+        match[1]
+      else
+        false
+    # Callback
+    (response) ->
+      fact = response.match
+      res.reply "#{fact} is #{facts[fact]}"
+  )
+```
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -84,7 +84,7 @@ In some situations, you want to dynamically match different messages (e.g. facto
 
 In a simple robot, this isn't much different from just putting the conditions in the Listener callback, but it makes a big difference when you are dealing with middleware: with the basic model, middleware will be executed for every match of the generic regex. With the dynamic matching model, middleware will only be executed when the dynamic conditions are matched.
 
-For example, the factoid lookup command could be reimplemented as:
+For example, the [factoid lookup command](https://github.com/github/hubot-scripts/blob/bd810f99f9394818a9dcc2ea3729427e4101b96d/src/scripts/factoid.coffee#L95-L99) could be reimplemented as:
 
 ```coffeescript
 module.exports = (robot) ->
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -264,14 +264,17 @@ module.exports = (robot) ->
 
 While the above helpers cover most of the functionality the average user needs (hear, respond, enter, leave, topic), sometimes you would like to have very specialized matching logic for listeners. If so, you can use `listen` to specify a custom match function instead of a regular expression.
 
+The match function must return a truthy value if the listener callback should be executed. The truthy return value of the match function is then passed to the callback as response.match.
+
 ```coffeescript
 module.exports = (robot) ->
   robot.listen(
-    (message) ->
-      # This function must return a truthy value if the listener callback should be executed.
-      # The return value of the match function is passed to the callback as response.match
-    (response) ->
-      # This is the standard listener callback
+    (message) -> # Match function
+      # Occassionally respond to things that Steve says
+      message.user.name is "Steve" and Math.random() > 0.8
+    (response) -> # Standard listener callback
+      # Let Steve know how happy you are that he exists
+      response.reply "HI STEVE! YOU'RE MY BEST FRIEND!"
   )
 ```
 
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -274,7 +274,7 @@ module.exports = (robot) ->
       message.user.name is "Steve" and Math.random() > 0.8
     (response) -> # Standard listener callback
       # Let Steve know how happy you are that he exists
-      response.reply "HI STEVE! YOU'RE MY BEST FRIEND!"
+      response.reply "HI STEVE! YOU'RE MY BEST FRIEND! (but only like #{response.match * 100}% of the time)"
   )
 ```
 
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -260,6 +260,21 @@ module.exports = (robot) ->
     res.send res.random leaveReplies
 ```
 
+## Custom Listeners
+
+While the above helpers cover most of the functionality the average user needs (hear, respond, enter, leave, topic), sometimes you would like to have very specialized matching logic for listeners. If so, you can use `listen` to specify a custom match function instead of a regular expression.
+
+```coffeescript
+module.exports = (robot) ->
+  robot.listen(
+    (message) ->
+      # This function must return a truthy value if the listener callback should be executed.
+      # The return value of the match function is passed to the callback as response.match
+    (response) ->
+      # This is the standard listener callback
+  )
+```
+
 ## Environment variables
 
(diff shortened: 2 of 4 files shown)


src/robot.coffee
```diff
@@ -68,6 +68,20 @@ class Robot
       @emit 'error', err
     process.on 'uncaughtException', @onUncaughtException
 
+  # Public: Adds a custom Listener with the provided matcher, options, and
```
michaelansel: By making this public, it does expose Listener.matcher as a new API. The API seems fairly well defined, but I'm not certain there are explicit test cases protecting it...

bhuga:
With this change, is this how you'd do something like #319?

```
robot.respondInRooms = (roomList, regex, options, callback) -> 
  matcher = (msg) -> msg.message.room.in roomList && msg.message.match(regex)
  robot.listen(matcher, options, callback)

robot.respondInRooms [1,2,3], /foo bar/, (response) ->
  response.reply "You just said foo bar in room 1, 2, or 3!"
```

I wonder if there's not a way to re-use TextListener here? I feel like having to rewrite the regex match means we're skipping a code path that catches a few corner cases.

michaelansel:
Yes to providing the capability for #319, in addition to #950 and #813. This could feasibly also replace the need for #871 and #868 (though those might still make sense to merge).

In its original incarnation (#871), I completely [removed `TextListener`](https://github.com/michaelansel/hubot/commit/e388efd8adc114ed346a394dcbb8720726892189) since it does [approximately nothing](https://github.com/michaelansel/hubot/blob/eavesdrop/src/robot.coffee#L95-L101).

I went ahead and left it out of scope here because of the potential complications with changing the listener class from `TextListener` to `Listener` and moving the regex to a listener option instead of an object property.

Overall, yes, I want to reduce duplication of regex matching code, but I feel like the volume of duplication is so low right now that I'd rather wait until we've seen how this will be used before creating an abstraction layer.

I like the example; I'll add it to the docs.

michaelansel:
Added an example to `patterns.md` along the lines of #813. My brain is a little :weary: right now though, so wordsmithing is welcome.

docs/scripting.md
```diff
@@ -260,6 +260,23 @@ module.exports = (robot) ->
...
 
+## Custom Listeners
+
+While the above helpers cover most of the functionality the average user needs (hear, respond, enter, leave, topic), sometimes you would like to have very specialized matching logic for listeners. If so, you can use `listen` to specify a custom match function instead of a regular expression.
+
+```coffeescript
+module.exports = (robot) ->
+  robot.listen(
+    (message) ->
+      # This function must return a truthy value if the listener callback should be executed.
```
technicalpickles: If you look at the other documentation in this file, it typically is a 'real' example, as in it's something you can copy and paste to run and it'd do something. It'd be good to have that here too, in addition to the comments.

technicalpickles:
I like the refactoring, and seems like a good way to expose a more generic listener. Still need to think on some of the implications though :grin:

src/robot.coffee
```diff
@@ -68,6 +68,20 @@ class Robot
...
+  # matcher  - A Function that determines whether to call the callback.
+  #            Expected to return a boolean.
+  # options  - An Object of additional parameters keyed on extension name
+  #            (optional).
+  # callback - A Function that is called with a Response object if the
+  #            matcher function returns true.
+  #
+  # Returns nothing.
+  listen: (matcher, options, callback) ->
+    @listeners.push new Listener(@, matcher, options, callback)
```
technicalpickles: For reference, the Listener constructor will grok out the right arguments if you pass it without options.

technicalpickles:
>In its original incarnation (#871), I completely removed TextListener since it does approximately nothing.
>
>I went ahead and left it out of scope here because of the potential complications with changing the listener class from TextListener to Listener and moving the regex to a listener option instead of an object property.

You might like some of the ideas in the vthree-rethink branch then. It removes the Listener subclasses:

* https://github.com/github/hubot/blob/vthree-rethink/src/robot.coffee#L58-L93
* https://github.com/github/hubot/blob/vthree-rethink/src/listener.coffee

bkeepers:
:+1: This looks good to me.

> Known issue: Robot.hear still creates TextListeners. Eventually, I'd like to get that moved over to just a standard Listener with special match logic, but need to decide what to do with the potential interface changes.

What do you think about deprecating TextListener now?

michaelansel:
Sounds like getting rid of `TextListener` eventually is the right decision. We should do the deprecation and removal separate from this PR though.

docs/scripting.md
michaelansel: Done

src/robot.coffee
michaelansel: :+1:

michaelansel:
Just closing the loop on the conversation, I think this is ready pending @technicalpickles's thinking about implications. Anything else that was on the list?

docs/patterns.md
```diff
@@ -77,3 +77,34 @@ module.export = (robot) ->
...
+    (message) ->
+      match = message.match(/^~(.*)$/)
+      # Only match if there is a matching factoid
+      if match and match[1] in facts
+        match[1]
+      else
+        false
+    # Callback
+    (response) ->
+      fact = response.match
```
technicalpickles: Does this work? I only see a `match` local variable in the matcher callback, nothing assigning it onto message object.

technicalpickles:
A few thoughts on implications:

* This exposes a new API, so people will be able to find and use it. Is it easy to use and understand? 
* There's some language overlap calling it `listen`, compared to `hear`. Is that confusing?

docs/patterns.md
michaelansel: `response.match` is set to the return value of the matcher function:

https://github.com/github/hubot/blob/master/src/listener.coffee#L39-L43
https://github.com/github/hubot/blob/master/src/response.coffee#L9

michaelansel:
> This exposes a new API, so people will be able to find and use it. Is it easy to use and understand?

Sort of? I feel like it is pretty simple and the docs explain everything special I can think of:
> The match function must return a truthy value if the listener callback should be executed. The truthy return value of the match function is then passed to the callback as `response.match`.

.
> There's some language overlap calling it `listen`, compared to `hear`. Is that confusing?

Erm... maybe? I don't think so... I think `listen` is definitely the right name though because it is a generic way of creating a Listener.

docs/patterns.md
```diff
@@ -77,3 +77,34 @@ module.export = (robot) ->
...
   robot.globalHttpOptions.httpsAgent = proxy('http://my-proxy-server.internal', true)
 ```
+
+## Dynamic matching of messages
+
+In some situations, you want to dynamically match different messages (e.g. factoids, JIRA projects). Rather than defining an overly broad regular expression that always matches, you can tell Hubot to only match when certain conditions are met.
+
+In a simple robot, this isn't much different from just putting the conditions in the Listener callback, but it makes a big difference when you are dealing with middleware: with the basic model, middleware will be executed for every match of the generic regex. With the dynamic matching model, middleware will only be executed when the dynamic conditions are matched.
+
+For example, the factoid lookup command could be reimplemented as:
```
technicalpickles: A link to this would be good, since a reader doesn't necessarily know how it's currently implemented. Also, might be including an abridged version here, in case it does get refactored to use this.

docs/scripting.md
```diff
@@ -260,6 +260,26 @@ module.exports = (robot) ->
     res.send res.random leaveReplies
 ```
 
+## Custom Listeners
+
+While the above helpers cover most of the functionality the average user needs (hear, respond, enter, leave, topic), sometimes you would like to have very specialized matching logic for listeners. If so, you can use `listen` to specify a custom match function instead of a regular expression.
+
+The match function must return a truthy value if the listener callback should be executed. The truthy return value of the match function is then passed to the callback as response.match.
```
technicalpickles: Might be worth mentioning what truthy is here or linking to something listing . Personally, I usually have to check if it includes blank string (that's falsey).

technicalpickles: It might be worth updating the example to actually make use of `response.match`.

src/robot.coffee
```diff
@@ -68,6 +68,20 @@ class Robot
       @emit 'error', err
     process.on 'uncaughtException', @onUncaughtException
 
+  # Public: Adds a custom Listener with the provided matcher, options, and
+  # callback
+  #
+  # matcher  - A Function that determines whether to call the callback.
+  #            Expected to return a boolean.
```
technicalpickles: It's not _just_ a boolean. It matches if it's truthy, and that truthy value is included as `response.match`

michaelansel: Good catch. Will update

docs/scripting.md
michaelansel: Hmm, okay. The example I had in mind (factoids) felt too complicated for `scripting.md`, so I put it in `patterns.md`. Can you think of a simpler example that would fit in `scripting.md`? Alternatively, I can just move the example over, but that would leave us example-less in `patterns.md`.

docs/patterns.md
michaelansel: K, I'll try to find something to link

docs/scripting.md
technicalpickles: How about something like:

```coffeescript
  robot.listen(
    (message) -> # Match function
      # Occasionally respond to things that Steve says
      if message.user.name is "Steve"
        steveRandomness = Math.random()
        if steveRandomness > 0.8
          steveRandomness

    (response) -> # Standard listener callback
      response.reply "HI STEVE! YOU'RE MY BEST FRIEND (but only like #{response.match * 100}% of the time)."
  )
```

michaelansel: Hahaha, I like it

michaelansel: I just realized that `message.user.name is "Steve" and Math.random() > 0.8` actually returns the random number (because it is the last thing in the expression), so `response.match` is already taken care of. Is just using that too clever for documentation?

michaelansel:
Okay, I think I addressed everything outstanding.

technicalpickles:
@michaelansel yep, covers everything :+1: 

Just a random thought after thinking about it in https://github.com/github/hubot/pull/996 is how adapters could leverage this to add custom listeners for custom message types. Slack attachments come to mind as one good use case.

PR #993, opened 2015-06-26
Add link to the Telegram Bot API adapter.

arcturial:
I included a link to a library I developed that allows interacting with [Telegram](https://telegram.org/) using the [Telegram Bot API](https://core.telegram.org/bots/api).

The library used to connect to Telegram currently supports all the available calls.

### DIFF
--- a/docs/adapters.md
+++ b/docs/adapters.md
@@ -40,6 +40,7 @@ to have yours added to the list:
 * [Slack](https://github.com/tinyspeck/hubot-slack)
 * [Skype](https://github.com/netpro2k/hubot-skype)
 * [Talker](https://github.com/unixcharles/hubot-talker)
+* [Telegram](https://github.com/lukefx/hubot-telegram)
 * [Twilio](https://github.com/jkarmel/hubot-twilio)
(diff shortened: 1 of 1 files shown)


lukefx:
Hey, same project here: https://github.com/lukefx/hubot-telegram
Why not collaborate on only one codebase?

arcturial:
I would be happy to collaborate, it was published around roughly the same time it seems. I would like to make use of the telegrambot adapter though to abstract the actual API interactions from the adapter code itself.

michaelansel:
When there is agreement on the canonical telegram adapter, I'll merge a PR pointing there. If there is disagreement, please update the PR to point to both adapters.

arcturial:
Thanks, I will have a discussion with @lukefx in a separate thread. Will update this PR as soon as it is resolved.

arcturial:
I have updated the link to point to https://github.com/lukefx/hubot-telegram where both authors code have been merged. You can now accept the PR if you are happy.

michaelansel:
Thanks for sorting that out!

PR #999, opened 2015-07-01
Add Skype Web to the list of adapters

sdimkov:


michaelansel:
Quick skim of the adapter code looks reasonable

PR #1003, opened 2015-07-04
Pass alias in Robot constructor. Fixes issue #1002.

### LINKED ISSUE #1002: Hubot doesn't know his alias during adapter initialization
Once Hubot constructor is called it initializes all fields and creates his adapter object. But one field is left behind to be configured later in runtime - Robot's alias. This prevents the adapter of knowing both robot's name and alias if it needs to. Also both name and alias can not be changed at runtime as they are hard-coded into every listener regex. So it makes no sense to leave the alias to be configured later.

I'll put a simple PR to fix this.

Also here is the use-case in which I hit this problem:
My adapter wants to know the respondPattern of the robot to be able to distinguish messages directed at the robot from the rest. For better efficiency the adapter calls respondPattern once from the constructor. But during that moment of time Hubot hasn't received his alias yet so I get incomplete regex from respondPattern.

sdimkov:
This PR enables passing alias to robot's constructor. It fixes #1002. It also does not disturb in any way any existing 3rd party code as it doesn't change the behavior of robot's constructor without the extra param being passed.

But why?
Because both Robot name and alias should be provided during creation of the robot. In the current design they can not be changed later and are hard-coded into every listener's regex. Therefore modifying alias or name after the robot is created is wrong and could cause problems.

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -84,13 +84,11 @@ if Options.create
 else
   adapterPath = Path.join __dirname, "..", "src", "adapters"
 
-  robot = Hubot.loadBot adapterPath, Options.adapter, Options.enableHttpd, Options.name
+  robot = Hubot.loadBot adapterPath, Options.adapter, Options.enableHttpd, Options.name, Options.alias
 
   if Options.version
     console.log robot.version
     process.exit 0
-
-  robot.alias = Options.alias
 
   loadScripts = ->
     scriptsPath = Path.resolve ".", "scripts"
--- a/index.coffee
+++ b/index.coffee
@@ -22,5 +22,5 @@ module.exports = {
   CatchAllMessage
 }
 
-module.exports.loadBot = (adapterPath, adapterName, enableHttpd, botName) ->
-  new Robot adapterPath, adapterName, enableHttpd, botName
+module.exports.loadBot = (adapterPath, adapterName, enableHttpd, botName, botAlias) ->
+  new Robot adapterPath, adapterName, enableHttpd, botName, botAlias
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -38,11 +38,11 @@ class Robot
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
-  constructor: (adapterPath, adapter, httpd, name = 'Hubot') ->
+  constructor: (adapterPath, adapter, httpd, name = 'Hubot', alias = false) ->
     @name      = name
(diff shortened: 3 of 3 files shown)


michaelansel:
Based on what I see, I like it. I'm not too familiar with this part of the code though. I'll try to find time to read through and gain context...

technicalpickles:
This makes sense to me.

One thought was about what else might be touching `alias`. It's usually checked for truthiness. Instead of a default `false`, it'd be getting undefined if you don't pass call it --alias but that's still falsy. That doesn't matter anyways since it's always getting assigned to `robot.alias` anyways, which would be undefined.

The other thought was that adding more parameters to the constructor, might be time to pass in an object rather than an array of parameters, but that can be a later refactoring.

sdimkov:
> it'd be getting undefined if you don't pass call it --alias

Nope :) That behavior is unchanged. You'd be getting `false` just as before.

That's because the new signature of the constructor is 
`constructor: (adapterPath, adapter, httpd, name = 'Hubot',  `**`alias = false`**`) ->`
Additionally the default value of `Options.alias` is also `false`

On passing object to the constructor: That's an excellent idea for refactoring but I have one concern:
This is a breaking change as it alters the way Robot objects are created. Are there any third parties that construct robots programmatically? I believe there are at least some PRs on the subject.

technicalpickles:
@sdimkov I was being a bit verbose to try to demostrate the type of things I look for in changes, but yeah, my conclusion was that it didn't change anything :grin: 

> Are there any third parties that construct robots programmatically? I believe there are at least some PRs on the subject.

They will be pretty few and far between. [hubot-mock-adapter](https://www.npmjs.com/package/hubot-mock-adapter) is the only one that comes to mind. Probably could do some trickery with checking the `typeof` the arguments, if the first one is an object, treat it as having everything, otherwise, treat it the way it is now and print a deprecation message. Could also poke at the arrity of the function call too.

sdimkov:
Okay, maybe we can merge this now ?

I don't mind writing the object-based constructor but that should be a separate PR.

sdimkov:
Guys, it's sad to see this still being open.

I don't understand why fixing bugs is such a low priority here. I personally believe that stability and polish are at this point far more important than adding a new feature X or Y. 

Is there anything more I can do to get this merged?

michaelansel:
Yeah, I know... We're all just a little short on free time :sweat:

I think this looks good as-is, with the one possibility of adding a test case for the default alias state. What do you think @technicalpickles?

PR #1004, opened 2015-07-06
Fix typo-s in robot class

sdimkov:


src/robot.coffee
```diff
@@ -392,7 +392,6 @@ class Robot
   # Returns nothing.
   parseHelp: (path) ->
     @logger.debug "Parsing help for #{path}"
-    scriptName = Path.basename(path).replace /\.(coffee|js)$/, ''
```
michaelansel: #917 actually depends on this being present. In addition, I'd much prefer to have test cases in place before modifying this code.

michaelansel:
Please remove the code change per my comments and I'll merge

sdimkov:
I've removed the commit. 

Are you sure #917 will get merged in the near future? It's open for the last 3 months

I agree with you in general about the importance of tests and coverage.Yet in this particular case that argument makes very little sense.

technicalpickles:
> Are you sure #917 will get merged in the near future?

I'm not really sure of it. I left my feedback down in https://github.com/github/hubot/pull/917#issuecomment-113532776 there.

PR #1006, opened 2015-07-09
Fix getting started links

jennifersmith:
Fixed a few broken links that I believe should point to docs/index.md

Note that I have not tested my changes as per [guidelines for contributing](CONTRIBUTING.md) as I just changed the text of an error message. Happy to do so if that is a problem ! Or you could just merge the other changes and I will be less lazy.

Thanks

Jen


sdimkov:
:+1:  Looks good to me. Thanks!

PR #1007, opened 2015-07-11
Update README.md

MyvarHD:


sdimkov:
Good idea :+1:  But you'll have to fix it before merging.

- You have a typo in the link: it points to `#hubote` channel.
- There is no need to put the same link twice.

MyvarHD:
all fixed.

PR #1012, opened 2015-07-24
clarify that a room token might be the expected parameter

alininja:
Hi,

Thanks for such an amazing project.  :)

I was testing out a hubot installation that uses the let's chat adapter.  So I'm not sure if this change only applies to the let's chat adapter or if it is applicable to all adapters.  That also means, I'm not sure if this is the correct place to update the docs.

Anyways, I couldn't get hubot to "hear" my curl request using just the chat room name.  Instead I needed to pass the chat room token to it.

Sorry, if this document change is way off.

Thanks again!

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -380,7 +380,7 @@ The most common use of this is for providing HTTP end points for services with w
 
 ```coffeescript
 module.exports = (robot) ->
-  robot.router.post '/hubot/chatsecrets/:room', (req, res) ->
+  robot.router.post '/hubot/chatsecrets/:room_name_or_token', (req, res) ->
     room   = req.params.room
     data   = if req.body.payload? then JSON.parse req.body.payload else req.body
     secret = data.secret
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -380,7 +380,8 @@ The most common use of this is for providing HTTP end points for services with w
 
 ```coffeescript
 module.exports = (robot) ->
-  robot.router.post '/hubot/chatsecrets/:room_name_or_token', (req, res) ->
+  # the expected value of :room is going to vary by adapter, it might be a numeric id, name, token, or some other value
+  robot.router.post '/hubot/chatsecrets/:room', (req, res) ->
     room   = req.params.room
     data   = if req.body.payload? then JSON.parse req.body.payload else req.body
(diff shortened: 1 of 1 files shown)


docs/scripting.md
```diff
@@ -380,7 +380,7 @@ The most common use of this is for providing HTTP end points for services with w
 
 ```coffeescript
 module.exports = (robot) ->
-  robot.router.post '/hubot/chatsecrets/:room', (req, res) ->
+  robot.router.post '/hubot/chatsecrets/:room_name_or_token', (req, res) ->
     room   = req.params.room
```
michaelansel: `req.params.room_name_or_token`

michaelansel:
Hmm... I see the confusion for chat services that have a unique ID for rooms that isn't the room name. I'm not sure whether this makes that much clearer though.

@bkeepers Do you have any good ideas on a better way of expressing that concept? IIRC Campfire does the same thing.

bkeepers:
Hmm, I'm not sure I understand what this is clarifying.  Currently, `req.params.room` can be either a name or a token, depending on the chat service, correct?

alininja:
Yes, room can be a name or a token.  The only thing I was trying to prevent was for new users to assume that room only means name.

Maybe instead of changing it from room to room_name_or_token, just a note letting users know that room can be a token?

Not sure, but I'm just pointing this out.

Thanks again! :smile:

docs/scripting.md
technicalpickles: What @michaelansel is pointing out is the route uses `room_name_or_token`, but then accesses `req.params.room`. It should be updated `req.params.room_name_or_token`.

technicalpickles:
Maybe just include a comment like "# the expected value of `room` is going to vary by adapter, it might be a numeric id, name, token, or some other value`.

technicalpickles:
@alininja one last thing... Your commits aren't attributed to your user. Can you work through https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/ to make sure your commits are attributed to your user? Once you are done, you should be able to work through https://help.github.com/articles/changing-author-info/ to update this branch's history.

alininja:
@technicalpickles it's okay if the commits aren't attributed to my user.  :smile:

PR #1019, opened 2015-08-05
Receive middleware

bhuga:
Builds off of #803 for middleware before all listeners. https://github.com/michaelansel/hubot/commit/b5bf215e44f20cf1c0c6080f4f2e19271f0e8242 was a bit too simple, but only just.

I need to test this a bit tomorrow but the tests look good.

cc @michaelansel @technicalpickles

### DIFF
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -88,26 +88,6 @@ There are a variety of different patterns for restricting access that you can fo
 * Specific access rules for every command (Role-based Access Control)
 * Blacklisting/whitelisting commands in specific rooms
 
-### Restricting access to all commands
-
-To completely prevent Hubot from operating for a given user, use a receive middleware:
-
-```
-
-BLACKLISTED_USERS = [
-  '12345' # Restrict access for a user ID for a contractor
-]
-
-robot.receiveMiddleware (context, next, done) ->
-  if context.response.message.user.id in BLACKLISTED_USERS
-    if context.response.message.text?.match(robot.respondPattern(''))
-      context.response.reply "I'm sorry @#{context.response.message.user.name}, but I'm configured to ignore your commands."
-    context.response.message.finish()
-    done()
-  else
-    next(done)
-```
-
 ### Simple per-listener access
 
 In some organizations, almost all employees are given the same level of access and only a select few need to be restricted (e.g. new hires, contractors, etc.). In this model, you partition the set of all listeners to separate the "power commands" from the "normal commands".
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -88,7 +88,27 @@ There are a variety of different patterns for restricting access that you can fo
 * Specific access rules for every command (Role-based Access Control)
 * Blacklisting/whitelisting commands in specific rooms
 
-### Simple (full vs restricted access)
+### Restricting access to all commands
+
+To completely prevent Hubot from operating for a given user, use a receive middleware:
+
+```
+
+BLACKLISTED_USERS = [
+  '12345' # Restrict access for a user ID for a contractor
+]
+
+robot.receiveMiddleware (context, next, done) ->
+  if context.response.message.user.id in BLACKLISTED_USERS
+    if context.response.message.text?.match(robot.respondPattern(''))
+      context.response.reply "I'm sorry @#{context.response.message.user.name}, but I'm configured to ignore your commands."
+    context.response.message.finish()
+    done()
+  else
+    next(done)
+```
+
+### Simple per-listener access
 
 In some organizations, almost all employees are given the same level of access and only a select few need to be restricted (e.g. new hires, contractors, etc.). In this model, you partition the set of all listeners to separate the "power commands" from the "normal commands".
 
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -622,11 +622,14 @@ These scoped identifiers allow you to externally specify new behaviors like:
 - authorization policy: "allow everyone in the `annoyers` group to execute `annoyance.*` commands"
 - rate limiting: "only allow executing `annoyance.start` once every 30 minutes"
 
-# Listener Middleware
+# Middleware
 
-Hubot supports inserting logic between the listener matching a message and the listener executing. This allows you to create extensions that apply to all scripts. Examples include centralized authorization policies, rate limiting, logging, and metrics. Middleware is implemented like other hubot scripts: instead of using the `hear` and `respond` methods, middleware is registered using `listenerMiddleware`.
+There are two kinds of middleware: Receive middleware and Listener Middleware.
 
-## Execution Process
+Receive middleware runs once, before listeners are checked.
+Listener middleware runs for every listener that matches the message.
+
+## Execution Process and API
 
 Similar to [Express middleware](http://expressjs.com/api.html#middleware), Hubot listener middleware executes middleware in definition order. Each middleware can either continue the chain (by calling `next`) or interrupt the chain (by calling `done`). If all middleware continues, the listener callback is executed and `done` is called. Middleware may wrap the `done` callback to allow executing code in the second half of the process (after the listener callback has been executed or a deeper piece of middleware has interrupted).
 
@@ -637,11 +640,26 @@ Middleware is called with:
   - response object (contains the original message)
 - next/done callbacks.
 
-For more details, see the [Listener Middleware API](#listener-middleware-api) section.
+- `context`
+  - See the each middleware type's API to see what the context will expose.
+- `next`
+  - a Function with no additional properties that should be called to continue on to the next piece of middleware/execute the Listener callback
+  - `next` should be called with a single, optional argument: either the provided `done` function or a new function that eventually calls `done`. If the argument is not given, the provided `done` will be assumed.
+- `done`
+ - a Function with no additional properties that should be called to interrupt middleware execution and begin executing the chain of completion functions.
+ - `done` should be called with no arguments
+
+Every middleware receives the same API signature of `context`, `next`, and
+`done`. Different kinds of middleware may receive different information in the
+`context` object. For more details, see the API for each type of middleware.
 
 ### Error Handling
 
 For synchronous middleware (never yields to the event loop), hubot will automatically catch errors and emit an an `error` event, just like in standard listeners. Hubot will also automatically call the most recent `done` callback to unwind the middleware stack. Asynchronous middleware should catch its own exceptions, emit an `error` event, and call `done`. Any uncaught exceptions will interrupt all execution of middleware completion callbacks.
+
+# Listener Middleware
+
+Listener middleware inserts logic between the listener matching a message and the listener executing. This allows you to create extensions that run for every matching script. Examples include centralized authorization policies, rate limiting, logging, and metrics. Middleware is implemented like other hubot scripts: instead of using the `hear` and `respond` methods, middleware is registered using `listenerMiddleware`.
 
 ## Listener Middleware Examples
 
@@ -698,24 +716,50 @@ module.exports = (robot) ->
 
 ## Listener Middleware API
 
-Although internal data structures are exposed, not all properties on the objects are considered part of the supported API. Below are the supported properties and usage information for each of the arguments middleware receives.
-
-- `context`
+Listener middleware callbacks receive three arguments, `context`, `next`, and
+`done`. See the [middleware API](#execution-process-and-api) for a description
+of `next` and `done`. Listener middleware context includes these fields:
   - `listener`
     - `options`: a simple Object containing options set when defining the listener. See [Listener Metadata](#listener-metadata).
     - all other properties should be considered internal
-
   - `response`
     - all parts of the standard response API are included in the middleware API. See [Send & Reply](#send--reply).
     - middleware may decorate (but not modify) the response object with additional information (e.g. add a property to `response.message.user` with a user's LDAP groups)
     - note: the textual message (`response.message.text`) should be considered immutable in listener middleware
 
-- `next`
-  - a Function with no additional properties that should be called to continue on to the next piece of middleware/execute the Listener callback
-  - `next` should be called with a single argument: either the provided `done` function or a new function that eventually calls `done`
+# Receive Middleware
 
-- `done`
- - a Function with no additional properties that should be called to interrupt middleware execution and begin executing the chain of completion functions.
- - `done` should be called with no arguments
+Receive middleware runs before any listeners have executed. It's suitable for
+blacklisting commands that have not been updated to add an ID, metrics, and more.
 
-Note that this API is protected by automated tests (`test/middleware_test.coffee`) to prevent breakage. Anything not covered by the tests is susceptible to change.
+## Receive Middleware Example
+
+This simple middlware bans hubot use by a particular user, including `hear`
+listeners. If the user attempts to run a command explicitly, it will return
+an error message.
+
+```coffeescript
+BLACKLISTED_USERS = [
+  '12345' # Restrict access for a user ID for a contractor
+]
+
+robot.receiveMiddleware (context, next, done) ->
+  if context.response.message.user.id in BLACKLISTED_USERS
+    if context.response.message.text?.match(robot.respondPattern(''))
+      context.response.reply "I'm sorry @#{context.response.message.user.name}, but I'm configured to ignore your commands."
+    context.response.message.finish()
+    done()
+  else
+    next(done)
+```
+
+## Receive Middleware API
+
+Receive middleware callbacks receive three arguments, `context`, `next`, and
+`done`. See the [middleware API](#execution-process-and-api) for a description
+of `next` and `done`. Receive middleware context includes these fields:
+  - `response`
+    - all parts of the standard response API are included in the middleware API. See [Send & Reply](#send--reply).
+    - middleware may decorate the response object with additional information (e.g. add a property to `response.message.user` with a user's LDAP groups)
+    - middleware may modify the `response.message` object
+    - note: the textual message (`response.message.text`) should be considered immutable in listener middleware
--- a/src/middleware.coffee
+++ b/src/middleware.coffee
@@ -21,7 +21,7 @@ class Middleware
   # Returns nothing
   # Returns before executing any middleware
   execute: (context, next, done) ->
-    done ||= ->
+    done ?= ->
     # Execute a single piece of middleware and update the completion callback
     # (each piece of middleware can wrap the 'done' callback with additional
     # logic).
--- a/src/middleware.coffee
+++ b/src/middleware.coffee
@@ -21,6 +21,7 @@ class Middleware
   # Returns nothing
   # Returns before executing any middleware
   execute: (context, next, done) ->
+    done ||= ->
     # Execute a single piece of middleware and update the completion callback
     # (each piece of middleware can wrap the 'done' callback with additional
     # logic).
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -51,6 +51,7 @@ class Robot
     @listeners  = []
     @middleware =
       listener: new Middleware(@)
+      receive:  new Middleware(@)
     @logger     = new Log process.env.HUBOT_LOG_LEVEL or 'info'
     @pingIntervalId = null
     @globalHttpOptions = {}
@@ -226,7 +227,7 @@ class Robot
   #
   # middleware - A function that determines whether or not a given matching
   #              Listener should be executed. The function is called with
-  #              (robot, listener, response, next, done). If execution should
+  #              (context, next, done). If execution should
   #              continue (next middleware, Listener callback), the middleware
   #              should call the 'next' function with 'done' as an argument.
   #              If not, the middleware should call the 'done' function with
@@ -237,7 +238,23 @@ class Robot
     @middleware.listener.register middleware
     return undefined
 
-  # Public: Passes the given message to any interested Listeners.
+  # Public: Registers new middleware for execution before matching
+  #
+  # middleware - A function that determines whether or not a given matching
+  #              Listener should be executed. The function is called with
+  #              (context, next, done). If execution should
+  #              continue to the next middleware or matching phase,
+  #              should call the 'next' function with 'done' as an argument.
+  #              If not, the middleware should call the 'done' function with
+  #              no arguments.
+  #
+  # Returns nothing.
+  receiveMiddleware: (middleware) ->
+    @middleware.receive.register middleware
+    return undefined
+
+  # Public: Passes the given message to any interested Listeners after running
+  #         receive middleware.
   #
   # message - A Message instance. Listeners can flag this message as 'done' to
   #           prevent further execution.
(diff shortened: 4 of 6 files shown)


test/robot_test.coffee
```diff
@@ -35,6 +35,7 @@ describe 'Robot', ->
 
     # Re-throw AssertionErrors for clearer test failures
     @robot.on 'error', (name, err, response) ->
+      return unless err?
```
bhuga: I managed to create some situations while testing that had an undefined `err` here. I'm not quite sure what I did, but the re-raiser was raising its own errors and it wasn't helpful :grin:

src/robot.coffee
```diff
@@ -247,6 +264,28 @@ class Robot
   # Returns nothing.
   # Returns before executing callback
   receive: (message, cb) ->
+    # When everything is finished (down the middleware stack and back up),
+    # pass control back to the robot
+    receiveMiddlewareDone = () ->
+      if cb?
+        process.nextTick -> cb true
```
bhuga: This little "make sure `done` is defined" trick is copied straight from `listener.coffee` from #803. I'll push it down to the middleware executor stage if there's agreement that we should.

src/robot.coffee
```diff
@@ -247,6 +264,28 @@ class Robot
...
+
+    @middleware.receive.execute(
+      {response: new Response(this, message)}
```
michaelansel: I don't think we should be creating a `Response` at this stage. The current model is a unique `Response` object for each matched `Listener`. I now see the problem there though: no `res.reply`, etc. Hmm....

src/robot.coffee
```diff
@@ -247,6 +264,28 @@ class Robot
...
+  # Private: Passes the given message to any interested Listeners.
+  #
+  # message - A Message instance. Listeners can flag this message as 'done' to
+  #           prevent further execution.
+  #
+  # cb - Optional callback that is called when message processing is complete
+  #
+  # Returns nothing.
+  # Returns before executing callback
+  processListeners: (message, context, done) ->
```
michaelansel: This is a middleware done callback; it should only be receiving `(context, done)` according to the middleware interface definition.

bhuga: Note that I'm sending a bound version of this :point_up:. It needs to be bound to `@robot` in any case (or in a wrapper function with `=>`), so I just bound `message` too.

src/robot.coffee
bhuga: `res.reply` is part of the problem, but it also creates an API that's inconsistent with listener middleware. We'd have to have `context.message` here and `context.response.message` there. It's a change, but I think creating a response here is the way to go.

bhuga:
This is working like I'd hope with this example:

```coffeescript
  robot.receiveMiddleware (context, next, done) ->
    if blacklistedUser(context.response) && !whitelistedCommand(context.response)
      if context.response.message.text?.match(/^(\/|hubot)/)
        context.response.reply "I'm sorry @#{context.response.message.user.name}, but I'm configured to ignore your commands."
      context.response.message.finish()
      done()
    else
      next(done)
```

src/robot.coffee
michaelansel: That's weird... the message should just be part of the context. That's the entire purpose of `context`

src/robot.coffee
michaelansel: Hmm... I want to think on this a little more. I don't agree...yet, but I'll keep thinking about it.

test/robot_test.coffee
technicalpickles: Hmm, the callback arity doesn't look is right. It should be `(err, response)`. `response` is optional, so maybe that's why it was failing? https://github.com/github/hubot/blob/receive-middleware/src/robot.coffee#L196-L202 for reference

src/robot.coffee
technicalpickles: Having a new `Response` seems okay to me. The only difference than other responses is that `response.match` will be undefined. I wouldn't expect a receive middleware to use that though, so it should be fine.

src/robot.coffee
technicalpickles: Seems okay to leave here for now. If there's another case that needs it in the executor stage, can refactor it then. (Would https://github.com/github/hubot/pull/1021 need it?)

test/robot_test.coffee
michaelansel: Hmmm.. I'm very confused by this. First, we have `robot.error != robot.on 'error'`, and then the funkiness around arity. I honestly don't remember what was going on, but I thought `name` was needed for some reason... Error handling probably needs a review pass and cleanup

src/robot.coffee
michaelansel: My concern is that this is a throw-away Response object: it will only be used by receive middleware, but not be available to the listener callback. This means, for example, that any attempts to inject custom response interception (response middleware) via receive middleware are going to fail.

I don't have a solution to `res.reply`, but I think that having a unique interface is okay. Each middleware stack serves a completely different purpose and will have different information available to it:
* Receive middleware: message
* Listener middleware: response, listener
* Response middleware: envelope

technicalpickles: > This means, for example, that any attempts to inject custom response interception (response middleware) via receive middleware are going to fail.

I'm not sure I follow what you mean.

src/robot.coffee
michaelansel: I think `Middleware.execute` should require a callback; anything that has an optional callback should take care of managing the optional-ness. Also, I don't think the `process.nextTick` is necessary here.

I would recommend changing this to:
```coffeescript
receiveMiddlewareDone = cb or ->
```

michaelansel:
Still trying to find time to do a full review of this (beyond just responding to specific comments)...

src/robot.coffee
michaelansel: My current idea for response middleware is to have it just be a piece of receive middleware instead of an actual change to the core. Similar to [how gzipping works in Express](https://github.com/expressjs/compression/blob/master/index.js#L60-L108), we would just monkey patch the response object to send responses through some proxy code. That would give us room to start playing with ideas without requiring core modifications or finalized APIs (i.e. faster iteration).

Another use case for receive middleware is automatically doing LDAP group lookups and attaching LDAP groups to something listener middleware and the listener callback can then utilize. While you can get the `message` object out and attach to that (because it passes through), that is strange and non-intuitive. As a consumer of the framework, I would expect either my entire object passes through or it doesn't. Having an identical object type (Response) that doesn't actually pass down is pretty weird.

technicalpickles: If we _don't_ have to access to a `response` object, then in order to send a message to the user/room that initiated, we'd need to do something like:

```coffeescript
envelope =
  room: context.message.room
  user: context.message.user
  message: context.message
robot.adapter.send envelop, "you've been denied"
```

The use case is having a receive middleware that blacklists specific users from using _any_ commands. So, it checks for the robot.respond regex, and replies they aren't trusted.

michaelansel: > The use case is having a receive middleware that blacklists specific users from using any commands. So, it checks for the robot.respond regex, and replies they aren't trusted.

I'm a little confused, do you mean something like this?

```coffeescript
robot.receiveMiddleware (context, next, done) ->
  if context.message.text.match robot.respondPattern /.*/
    # this is a message for the the bot
    if userInBlacklist context.message.user
      # don't let them do anything
    else
      # all good!
      next done
  else
    # not a message for the bot
    next done
```

Assuming yes, I somewhat disagree with that. I feel like that is what listener middleware is for. Middleware shouldn't be doing pattern matching on message text; that is the job of a Listener. That means that you should only be blocking actual commands and not just responding to arbitrary messages using middleware.

Does that make sense? If not, I'll try to explain that again a different way.

michaelansel: Just saw the example from below. Yeah, I really think listener middleware is the correct place for that kind of blacklisting. While it could be accomplished using receive middleware, it should be considered a hack (though the whole point of middleware is to allow this kind of customization).

Drawing parallels to Express: in Express, you have a single request/response pair. When a request comes in, a response object is generated to handle responding to that request. In Hubot, we have multiple "requests": one request per matched listener. As a result, there is one response object for handling the response from that listener.

We could change the model so that a message is considered a request and there is a single response object shared across all matching listeners. Alternatively, we could decide that the Express request/response model doesn't work well for us and define something completely different.

michaelansel: > This means, for example, that any attempts to inject custom response interception (response middleware) via receive middleware are going to fail.

And I just realized this statement is completely stupid. A throwaway Response object is no different from not having a Response object with respect to response middleware. Nevermind...

test/robot_test.coffee
bhuga: > Error handling probably needs a review pass and cleanup

I'm going to leave this one little guard in place for now, then.

src/robot.coffee
bhuga: > We could change the model so that a message is considered a request and there is a single response object shared across all matching listeners.

This results in weirdness inside listeners. `response.match` will be unique to each listener, but `response.message` would be shared. It also means that an early listener can change a message to no longer match another listener. I'm not sure that's an interaction I like; I feel like message modification that affects listeners should be the job of a receive middleware.

>  Alternatively, we could decide that the Express request/response model doesn't work well for us and define something completely different.

In chat, one "request" can fire multiple listeners and create multiple responses, a pretty different model than how HTTP works. Express isn't bad, and if we can be consistent, that's good. But I have a feeling that API will feel pretty awkward.

src/robot.coffee
bhuga: Fixed this (it still binds to `this`, but uses the context's copy of the message now).

src/robot.coffee
bhuga: Adopted.

bhuga: > Seems okay to leave here for now. If there's another case that needs it in the executor stage, can refactor it then. (Would #1021 need it?)

#1021 also drops an empty `done` here: 
https://github.com/github/hubot/pull/1021/files#diff-25213c38e7e2c1e60eda9b1c1c473c60R75

bhuga: https://github.com/github/hubot/commit/20fce05297bf821063863307433f3fb5a7f1926a adds a default `done`, and I like it better. It will fix #1021 transparently.

src/middleware.coffee
```diff
@@ -21,6 +21,7 @@ class Middleware
   # Returns nothing
   # Returns before executing any middleware
   execute: (context, next, done) ->
+    done ||= ->
```
technicalpickles: Shouldn't this be `?=`?

bhuga: I don't think so, but this might be the extent of my coffeescript knowledge. `?=` would let us pass in `0` or 100 other falsey things and have them be silently overwritten. `||=` would only fire in case of a skipped argument.

technicalpickles: I thought it was a rubyism for some reason.

From http://coffeescript.org/#operators:

> It's a little difficult to check for the existence of a variable in JavaScript. if (variable) ... comes close, but fails for zero, the empty string, and false. CoffeeScript's existential operator ? returns true unless a variable is null or undefined, which makes it analogous to Ruby's nil?
>
>It can also be used for safer conditional assignment than ||= provides, for cases where you may be handling numbers or strings.

src/robot.coffee
```diff
@@ -226,7 +227,7 @@ class Robot
   #
   # middleware - A function that determines whether or not a given matching
   #              Listener should be executed. The function is called with
-  #              (robot, listener, response, next, done). If execution should
+  #              (context, next, done). If execution should
```
technicalpickles: I was wondering about this change, but i'm pretty sure it was just a stale doc from before we had a context object.

bhuga: Yep, that's exactly what it is.

src/robot.coffee
```diff
@@ -247,6 +264,24 @@ class Robot
...
+    )
+
+  # Private: Passes the given message to any interested Listeners.
+  #
+  # message - A Message instance. Listeners can flag this message as 'done' to
+  #           prevent further execution.
+  #
+  # cb - Optional callback that is called when message processing is complete
+  #
+  # Returns nothing.
```
technicalpickles: I see a few other places where we `return undefined` explicitly. I think otherwise, it'd return whatever async.detectServies` returns (maybe that's nothing, hard to tell from the code)

bhuga: Great catch, explicit return added.

src/robot.coffee
michaelansel: I agree. I misremembered what [async does](https://github.com/caolan/async#seriestasks-callback) and thought callbacks were required. Turns out optional callbacks is a normal thing.

src/middleware.coffee
michaelansel: :+1: to `?=`

src/robot.coffee
```diff
@@ -237,7 +238,23 @@ class Robot
     @middleware.listener.register middleware
     return undefined
 
-  # Public: Passes the given message to any interested Listeners.
+  # Public: Registers new middleware for execution before matching
+  #
+  # middleware - A function that determines whether or not a given matching
+  #              Listener should be executed. The function is called with
```
michaelansel: "A function that determines whether or not Listener matching should commence for a given Message." (or something like that)

src/robot.coffee
```diff
@@ -247,6 +264,24 @@ class Robot
...
+      @processListeners.bind this
+      cb
+    )
+
+  # Private: Passes the given message to any interested Listeners.
+  #
+  # message - A Message instance. Listeners can flag this message as 'done' to
+  #           prevent further execution.
+  #
+  # cb - Optional callback that is called when message processing is complete
```
michaelansel: Looks like you changed this to `done` instead of `cb`

test/middleware_test.coffee
```diff
@@ -301,7 +309,11 @@ describe 'Middleware', ->
       @robot.server.close()
       @robot.shutdown()
 
-    describe 'context', ->
+    describe 'listner middleware context', ->
```
michaelansel: listEner

test/robot_test.coffee
michaelansel: I guess I ran into this before, but solved [a slightly different way](https://github.com/github/hubot/blob/receive-middleware/test/middleware_test.coffee#L291). I think it will handle an additional edge case, but I don't know how much it matters. I have absolutely no memory of that though, so who knows.

test/middleware_test.coffee
```diff
@@ -337,7 +349,29 @@ describe 'Middleware', ->
...
+                  sinon.match.has('message',
+                    sinon.match.same(@testMessage)))) # context
+              sinon.match.any                         # next
+              sinon.match.any                         # done
+            )
+            testDone()
+
     describe 'next', ->
+      beforeEach ->
+        @robot.listenerMiddleware (context, next, done) =>
```
michaelansel: Do we care that there is no test coverage of `next`/`done` for receive middleware? I *think* it might be okay, but my brain isn't 100% here today.

bhuga: I saw this as an integration test, and worth having, but I think the robot tests are the right place to have full coverage for the different kinds of middleware (and I added a bunch of tests there).

test/robot_test.coffee
bhuga: I changed it up to check the constructor as well, but I do always try to reduce nesting with early returns.

src/middleware.coffee
bhuga: `?=` it is, adopted.

bhuga:
Good :eyes:, little cleanups added.

michaelansel:
Overall, I like this. The only sticking point for me is around what goes in the context object (Response vs Message).

Response
* Pro: Easy access to `res.reply` for message sending
* Con: Create an incomplete Response (no `match`) for exclusively middleware execution (thrown away afterwards)
* Con: Feels weird that you can only modify `res.message` and have it propagate.

Message
* Pro: Feels "right". This is pre-Listener, so Responses shouldn't exist yet.
* Pro: Easy and obvious to modify the Message object and have changes propagate on to Listeners.
* Con: Sending messages is annoyingly complicated.

Is that a fair comparison of the options? I would really rather focus on making message sending easier and keep the middleware interface clean. Maybe something like `robot.reply msg, "I see you said something"` and it takes care of envelope construction. Or, maybe we say that sending replies from receive middleware is unsupported (I have no idea of the implications here).

bhuga:
> Is that a fair comparison of the options?

Mostly, but I'm not sure I agree with this one:

> Pro: Feels "right". This is pre-Listener, so Responses shouldn't exist yet.

Responses are where replies come from. This doesn't feel so unnatural to me.

I do share your unease that `context.response.message` is something that can be edited; it's buried too deep in the object to feel first-class that way. And it's very icky that a response has a match, and sometimes not. But I don't think there's anything inherently wrong with a `Response` happening before listeners have run; we're still *responding to a message,* after all.

> I would really rather focus on making message sending easier and keep the middleware interface clean. Maybe something like robot.reply msg, "I see you said something" and it takes care of envelope construction.

My gut reaction is a pretty strong "ick" to this. Being given the message so that I can pass it to some other function just feels awkward, like I'm doing work that the context could have set up for me. We could have `context.reply()`, for example, which would be cleaner and wrap the message up.

Even if we get past that, I'm also nervous that anywhere we put `reply` we have to add `send`, `emote`, `topic`, etc. This is a messy API already on `Response`, and I wouldn't want to make another object that replicates it.

> Or, maybe we say that sending replies from receive middleware is unsupported (I have no idea of the implications here).

I think this is a non-starter if only for error messages.

technicalpickles:
>> Pro: Feels "right". This is pre-Listener, so Responses shouldn't exist yet.
>Responses are where replies come from. This doesn't feel so unnatural to me.
>
> .... But I don't think there's anything inherently wrong with a Response happening before listeners have run; we're still responding to a message, after all.

That's my thinking too :+1: 

> Even if we get past that, I'm also nervous that anywhere we put reply we have to add send, emote, topic, etc. This is a messy API already on Response, and I wouldn't want to make another object that replicates it.

I'm worried about that too. These methods ultimately delegate to the adapter, so anytime there's adapter changes, it's more places to update.

> I do share your unease that context.response.message is something that can be edited; it's buried too deep in the object to feel first-class that way.

Saying it that way, makes me think `context.message` should be documented as the thing to modify if you want to. If we wanted to be really fancy, we could probably check if context.response.message was adjusted at all, and give a warning.

michaelansel:
Okay, I think I'm sufficiently convinced this is the most straightforward solution. I'd like to revisit it if we ever do a major rev on the adapter interface, but I think it will work well for the foreseeable future.

:+1:, let's do it!

bhuga:
I think the only comment open here is:

> Saying it that way, makes me think context.message should be documented as the thing to modify if you want to. If we wanted to be really fancy, we could probably check if context.response.message was adjusted at all, and give a warning.

I think we should just leave it as `context.response.message`. Having the message be accessible in two places leaves the user to wonder if there's a difference; the natural instinct is to make them be the very same object. And if we fight that, we end up with disposable, cloned message objects, and users wondering which one is correct to change unless we add the aforementioned change-detection machinery.

I think simplicity is the better part of valor here, and if you agree, we can land this :sparkles:

technicalpickles:
@bhuga that makes sense to me. As long the documentation talks about that message as being modifyable, such that it'd affect other receive middleware and listeners, should be good.

Speaking of documentation, guess what this PR needs? :grin:

bhuga:
> Speaking of documentation, guess what this PR needs?

This PR needs a new hobby so that it can feel *fulfilled* in life. It's considering taking up gardening. But in lieu of that, I've added documentation. It required reworking the docs a bit, since some of the stuff that applies to all middleware. :eyes: whenever.

> As long the documentation talks about that message as being modifyable, such that it'd affect other receive middleware and listeners, should be good.

I :books:'d this and added a test, too.

docs/patterns.md
```diff
@@ -88,7 +88,27 @@ There are a variety of different patterns for restricting access that you can fo
 * Specific access rules for every command (Role-based Access Control)
 * Blacklisting/whitelisting commands in specific rooms
 
-### Simple (full vs restricted access)
+### Restricting access to all commands
+
+To completely prevent Hubot from operating for a given user, use a receive middleware:
```
technicalpickles: Isn't the same exact example in the scripting documentation?

bhuga: It was, as I think it's the most basic receive middleware, but you're right that it's weird to duplicate it. On further thought, I removed it from the patterns document. I'm not sure it's a popular enough use case to call a "pattern"; hubot's gone years without it, after all.

docs/scripting.md
```diff
@@ -698,24 +716,50 @@ module.exports = (robot) ->
...
 
-- `done`
- - a Function with no additional properties that should be called to interrupt middleware execution and begin executing the chain of completion functions.
- - `done` should be called with no arguments
+Receive middleware runs before any listeners have executed. It's suitable for
+blacklisting commands that have not been updated to add an ID, metrics, and more.
+
+## Receive Middleware Example
+
+This simple middlware bans hubot use by a particular user, including `hear`
```
michaelansel: Wait, that's not true (unless I'm misreading). The code below looks for anything that matches the respond pattern. If you want to block hear listeners, you need to use listener middleware (only way to know if a message will match).

docs/scripting.md
```diff
@@ -698,24 +716,50 @@ module.exports = (robot) ->
...
+## Receive Middleware API
+
+Receive middleware callbacks receive three arguments, `context`, `next`, and
+`done`. See the [middleware API](#execution-process-and-api) for a description
+of `next` and `done`. Receive middleware context includes these fields:
+  - `response`
+    - all parts of the standard response API are included in the middleware API. See [Send & Reply](#send--reply).
+    - middleware may decorate the response object with additional information (e.g. add a property to `response.message.user` with a user's LDAP groups)
+    - middleware may modify the `response.message` object
+    - note: the textual message (`response.message.text`) should be considered immutable in listener middleware
```
michaelansel: lolwhoops. Those two bullet points contradict each other.

docs/scripting.md
bhuga: The `responsePattern` checks if someone was trying to use a command and returns an error if so. The message is always ignored, even `.hear`, it's just a matter of if we tell the sender so.

docs/scripting.md
bhuga: copypasta. I will fix this.

docs/scripting.md
michaelansel: I'm confused. Do you mean `robot.hear /^hubot: stuff$/` would get ignored? Yes, I agree, but that isn't how the docs read. I would interpret the docs as "This will block `robot.hear /hello/`", but that is false.

bhuga: > Yes, I agree, but that isn't how the docs read. I would interpret the docs as "This will block robot.hear /hello/", but that is false.

Check the nesting; this blocks all messages, and only responds with an error message on "hubot foo" or "/foo".

I will readjust this code so the always-blocking behavior is front and center.

michaelansel: Ohhhh, now I understand. I didn't read the code closely enough. Thanks for clarifying!

bhuga: No worries, in retrospect having the code after an indent like that is pretty fugly. I've reworked it in #1038.

PR #1021, opened 2015-08-06
Response middleware

bhuga:
Hot on the heels of #1019, response middleware gives you the chance to edit outgoing messages.

It looks like this:

```coffeescript
# Replace markdown links with slack's <url|text> format.
module.exports = (robot) ->
  robot.responseMiddleware (context, next, done) ->
    regex = /\[([^\[\]]*?)\]\((https?:\/\/.*?)\)/g

    context.string = context.string.replace(regex, "<$2|$1>")
    next(done)
```

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -644,21 +644,17 @@ These scoped identifiers allow you to externally specify new behaviors like:
 
 # Middleware
 
-There are two kinds of middleware: Receive middleware and Listener Middleware.
+There are three kinds of middleware: Receive, Listener and Response.
 
 Receive middleware runs once, before listeners are checked.
 Listener middleware runs for every listener that matches the message.
+Response middleware runs for every response sent to a message.
 
 ## Execution Process and API
 
-Similar to [Express middleware](http://expressjs.com/api.html#middleware), Hubot listener middleware executes middleware in definition order. Each middleware can either continue the chain (by calling `next`) or interrupt the chain (by calling `done`). If all middleware continues, the listener callback is executed and `done` is called. Middleware may wrap the `done` callback to allow executing code in the second half of the process (after the listener callback has been executed or a deeper piece of middleware has interrupted).
+Similar to [Express middleware](http://expressjs.com/api.html#middleware), Hubot executes middleware in definition order. Each middleware can either continue the chain (by calling `next`) or interrupt the chain (by calling `done`). If all middleware continues, the listener callback is executed and `done` is called. Middleware may wrap the `done` callback to allow executing code in the second half of the process (after the listener callback has been executed or a deeper piece of middleware has interrupted).
 
 Middleware is called with:
-
-- a context object containing:
-  - matching Listener object (with associated metadata)
-  - response object (contains the original message)
-- next/done callbacks.
 
 - `context`
   - See the each middleware type's API to see what the context will expose.
@@ -788,3 +784,31 @@ of `next` and `done`. Receive middleware context includes these fields:
     - this response object will not have a `match` property, as no listeners have been run yet to match it.
     - middleware may decorate the response object with additional information (e.g. add a property to `response.message.user` with a user's LDAP groups)
     - middleware may modify the `response.message` object
+
+# Response Middleware
+
+Response middleware runs against every message hubot sends to a chat room. It's
+helpful for message formatting, preventing password leaks, metrics, and more.
+
+## Response Middleware Example
+
+This simple example changes the format of links sent to a chat room from
+markdown links (like [example](https://example.com)) to the format supported
+by [Slack](https://slack.com), <https://example.com|example>.
+
+```coffeescript
+module.exports = (robot) ->
+  robot.responseMiddleware (context, next, done) ->
+    context.strings = (string.replace(/\[([^\[\]]*?)\]\((https?:\/\/.*?)\)/, "<$2|$1>") for string in context.strings)
+    next()
+```
+
+## Response Middleware API
+
+Response middleware callbacks receive three arguments, `context`, `next`, and
+`done`. See the [middleware API](#execution-process-and-api) for a description
+of `next` and `done`. Receive middleware context includes these fields:
+  - `response`
+    - This response object can be used to send new messages from the middleware. Middleware will be called on these new responses. Be careful not to create infinite loops.
+  - `strings`
+    - An array of strings being sent to the chat room adapter. You can edit these, or use `context.strings = ["new strings"]` to replace them.
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -800,6 +800,7 @@ by [Slack](https://slack.com), <https://example.com|example>.
 ```coffeescript
 module.exports = (robot) ->
   robot.responseMiddleware (context, next, done) ->
+    return unless context.plaintext?
     context.strings = (string.replace(/\[([^\[\]]*?)\]\((https?:\/\/.*?)\)/, "<$2|$1>") for string in context.strings)
     next()
 ```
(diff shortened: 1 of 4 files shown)


technicalpickles:
By calling `robot.adapter.whatever` multiple times, rather than passing it an array of `strings...`, I have a feeling this is going to break some things. The thing that comes to mind is calling `response.send` multiple times causing messages to come out in random order for the campfire adapter, where `response.send` with multiple arguments comes out in the correct order.

bhuga:
Okay, with #1019 landed, I picked this back up.

> By calling robot.adapter.whatever multiple times, rather than passing it an array of strings..., I have a feeling this is going to break some things.

I adjusted this to send the list of strings to each middleware instead of each string. It's slightly more cumbersome on the middleware, but 100% backwards compatible.

I also added some docs and a couple of tests. Ready for :eyes: again.

michaelansel:
We probably also need a suite of tests for message sending pre-middleware. Not much there right now to protect the existing state.

bhuga:
> We probably also need a suite of tests for message sending pre-middleware. Not much there right now to protect the existing state.

This is tough, because hubot-mock-adapter does not actually obey the `string, string, string, function()` semantics (though now that I look at it, neither does the shell adapter). I'm afraid the different adapter behaviors make the current behavior not terribly well-defined.

michaelansel:
:sob: Okay, I guess we just leave the Response/Adapter interface as it is right now (a crazy mess) and put our hopes and dreams on #1036.

bhuga:
> Okay, I guess we just leave the Response/Adapter interface as it is right now (a crazy mess) and put our hopes and dreams on #1036.

Singing songs of #1036...

Would this interfere with that? I don't think it would; can we move forward on this separately?

michaelansel:
Yeah, I think this can move forward separately. As long as this is bound to the normal response interface (all the existing function calls), we can ensure backwards compatibility with all manner of shim layers in the core. It ain't gonna be pretty, but it does allow us to segment out the decision making.

technicalpickles:
As an author of response middleware, do you think you'd need to be able to target specific Response methods? For example, would you want to be modify topics vs send and reply differently? I suspect you would, so that would be a good thing to include in the `context`.

bhuga:
> I suspect you would, so that would be a good thing to include in the context.

Really good point! Do you think just `context.method` would be okay?

bhuga:
I added `context.method` as the string representation of the method, so `send`, `reply`, etc.

michaelansel:
How does this look after #1036 merges and we port all the existing methods over to a singular flow (similar to how Message encompasses all kinds of input)?

michaelansel:
My first reaction is concern, but I *think* this will be okay when mixed in with #1036 as long as the `context.method` value doesn't change. This has a similar feel to #949 (using named object types instead of inheritance). Without spending time thinking through the long-term design of #1036, I'm not really sure how this will fit, but I *think* it is okay???

@bhuga, can you try to rough out what this would look like when merged with the ideas in #1036? (i.e. custom prompts, `sendImage`, `sendSticker`, etc.) Should we make the response message types more abstract? (e.g. send/reply are both text responses, just like hear/respond are both text listeners)

#michaelneedsmoretimeforthis :sob:

bhuga:
> @bhuga, can you try to rough out what this would look like when merged with the ideas in #1036? (i.e. custom prompts, sendImage, sendSticker, etc.) Should we make the response message types more abstract? (e.g. send/reply are both text responses, just like hear/respond are both text listeners)

I'll noodle on this. We're running into this sort of thing too, and it's plain that a pile of `string`s is not going to cut it.

We should also decide if the callback as the last argument to `send` and friends is something we support or not.

bhuga:
> @bhuga, can you try to rough out what this would look like when merged with the ideas in #1036? (i.e. custom prompts, sendImage, sendSticker, etc.) Should we make the response message types more abstract? (e.g. send/reply are both text responses, just like hear/respond are both text listeners)

I thought on this some. I don't think we should make them more abstract. My term `strings` is wrong.

Different chat systems use structured data to handle messages. I don't think it's possible for a generic middleware to understand all of this structured data, or to provide an interface to make it "plaintext enough". If a script returns structured data, we pass that through to this middleware. So I would rename `strings` to, say, `messages`, but I think it's still fine to have a `method` and a list of data objects.

But I don't see #1036 coming up with something that wouldn't be able to fit that pattern. The method might become `keyboard` for telegram, and middleware might receive structured data and need to know how to handle it. But I think this response middleware interface would work.

bhuga:
Any objections to merging? Would love to get starting using this.

technicalpickles:
I was thinking about how to deal with adapter's having different maximum lengths of message they support sending, and it occurred to me that response middleware could be useful for that.

Currently, if you send a large message with most adapters, they silently truncate, or the service itself truncates on their behalf. As a developer of a hubot, you wouldn't know that. I was thinking one approach to that would be to have a response middleware that checks the length of a message, and it could do something like emit an error so the developers could track down what was generating large messages.

docs/scripting.md
```diff
@@ -788,3 +784,33 @@ of `next` and `done`. Receive middleware context includes these fields:
...
+
+Response middleware callbacks receive three arguments, `context`, `next`, and
+`done`. See the [middleware API](#execution-process-and-api) for a description
+of `next` and `done`. Receive middleware context includes these fields:
+  - `response`
+    - This response object can be used to send new messages from the middleware. Middleware will be called on these new responses. Be careful not to create infinite loops.
+  - `strings`
+    - An array of strings being sent to the chat room adapter. You can edit these, or use `context.strings = ["new strings"]` to replace them.
+  - `method`
+    - A string representing which type of response message the listener sent, such as `send`, `reply`, `emote` or `topic`.
```
michaelansel: Should this instead be a message type instead of the method name? I think it is better to specify `if message.type == text` instead of `if message.type in ['send','reply'], but that depends on whether we want to allow differentiation between those types of messages... Maybe not worth worrying about?

bhuga: Good thought, I hadn't considered that.

Types might be better in some cases, but the edges are fuzzier. Internally we're abstracting away colors in slack attachments, for example, and I would want to consider a helper for that as "text". But is an emote text? Is a topic? What type would a slack attachment, a sent file, or telegram buttons be?

My first instinct is that `type` would be fuzzy for some of the more basic messages, and have a 1:1 correlation with advanced message formats in specific adapters. It's thus worse than `send` in the base case and gains no value in the adapter-specific case, instead being a piece of metadata boilerplate that custom message types require without much benefit.

But I'm not really sure. Can you think of any complex messages that could safely share the same type?

MrSaints:
I wrote a really simple, and [hacky response middleware](https://github.com/ClaudeBot/hubot/commit/0eec34462f712f62b2fe3d679784f5f585d27740) a while back.

I much prefer this PR though. I'm really glad this is being done though, and hopefully it'll be merged soon :+1:

docs/scripting.md
michaelansel: I think you're right here. I want there to be a way to group and abstract outgoing messages, but maybe it isn't so easy. I'm really just squeamish about seeing this everywhere: `if message.type in ['send','reply']`. Can you think of another way to make that extremely common case less gross?

bhuga: Hm. `message.plaintext`? Adapters could set it alongside basic `send` and `reply`, and it arguably applies to `topic` and `emote`, too.

michaelansel: Missed this, but independently had the exact same idea. Thinking in the same style as Ruby `message.plaintext?` returns a boolean. No `?` in CS though, so I like just having it as a "read-only" property. Thinking about applying the same idea to inbound messages to make `robot.listen` easier to work with (instead of needing to do `message instanceof TextMessage` every time)

bhuga: I adopted the `context.plaintext` pattern. See what you think. https://github.com/github/hubot/compare/f0acd0f6e8cc240af1ee687da22071f5c993e4d9...0544029fb3f8d0964228cf5f379e160bea48b573 has the diff from the last time you looked until now.

michaelansel:
:+1: I like it. I think we can probably merge this.

bhuga:
The triumvirate of middleware is complete. Half life 3 confirmed.

Issue #873, opened 2015-02-26
New Script Pattern


michaelansel:
I had a crazy pipe dream idea for a new script format that I think will still be backwards-compatible, be fairly powerful, and not require programmatic launching (#858). There is a lot in this example, so I've added lots of comments and will go through each piece. In short, it probably works as you would expect it to.

``` coffeescript
# this replaces module.exports = (robot) ->
class Sarcasm extends HubotScript
  constructor: (@robot) ->
    super @robot, 'sarcasm' # give every script a unique name

  # Specify config requirements
  config:
    'HUBOT_SARCASM_CLEAN':
      required: true # require that this config option is set
      description: "Keep everything safe for work" # used to generate help/error messages
      validation: (value) -> # maybe a function to parse/validate the input?
    'HUBOT_SARCASM_LEVEL':
      required: false # optional (feels a little redundant with default)
      default: 7 # default to 7
      description: "How sarcastic should hubot be? (1-7)"

  # Specify dependencies
  dependencies:
    capabilities:
      'private-message' # require an adapter that supports sending direct/private messages
    middleware:
      'ldap-user-info' # need this middleware to be loaded

  # Data for a theoretical extension that runs automated integration tests against all loaded scripts
  examples:
    # say something and then expect to the pattern to be matched
    "I'm just saying hello...": /sup/
    "Hubot: orly": /^User: rly rly$/

  # Specify all routes/listeners
  routes:
    # equivalent to:  robot.hear /hello/, (msg) -> msg.say 'sup'
    'hello':
      command: true # injects the "respond" prefix
      pattern: /hello/
      handler: (response) ->
        response.say 'sup'

    'orly':
      command: false
      help:
        'Hubot: orly': "Troll anyone who says 'orly' to Hubot"
      pattern: /orly/
      handler: (response, done) -> # arity == 2 causes this to be treated as an async function
        @doSomeAwesome response, done

    'say-hi-to-new-people':
      event: 'enter'
      handler: (response) ->
        response.send 'hi new person!'

    'sass-at-new-topics':
      event: 'topic'
      handler: (response) ->
        response.reply 'orly?'

  # Helper functions (that are now easily accessible for unit testing!)
  doSomeAwesome: (response, cb) ->
    process.nextTick ->
      response.reply 'rly rly'
      cb()

module.exports = Sarcasm
```

[...]

PR #1025, opened 2015-08-11
makes script-documentation accessible via the robot.scriptDocumentati…

sharabash:
…on instance property

The parseHelp method already stores scriptDocumentation in a hash but does absolutely nothing with it. This mod pushes the stored documentation to an accessible @scriptDocumentation property so that other modules/scripts can make use of it (see https://www.npmjs.com/package/hubot-advanced-help, for example, which is a module I wrote that assumes this property exists). In the absence of this property, I'm going to have to resort to opening sibling scripts and parsing in the documentation even though the robot.coffee script already does that (it just does nothing with the data).

cc @imperialwicket @nealfax

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -54,6 +54,7 @@ class Robot
     @logger     = new Log process.env.HUBOT_LOG_LEVEL or 'info'
     @pingIntervalId = null
     @globalHttpOptions = {}
+    @scriptDocumentation = []
 
     @parseVersion()
     if httpd
@@ -466,6 +467,8 @@ class Robot
         cleanedLine = line[2..line.length].replace(/^hubot/i, @name).trim()
         scriptDocumentation.commands.push cleanedLine
         @commands.push cleanedLine
+
+    @scriptDocumentation.push scriptDocumentation
 
   # Public: A helper send function which delegates to the adapter's send
   # function.


michaelansel:
I like the goal here. If we are going to go down this route, I would really like to combine with #917 and get a single, nice data structure that contains all the documentation separated by script. I'm game for iteration, but the first step needs to be hiding the internal data structures and exposing query methods instead.

technicalpickles:
@sharabash thanks for fixing up the commit authorship!

Two other pieces of feedback:

* do you think it'd be more useful for accessing script documentation by script name (ie use an object instead of an array)?
* could you add documentation to `doc/scripting.md` so script authors can discover it? I try to add documentation for these type of things, so it's not accidentally removed from core when it looks like it's not being used anything.

michaelansel:
How about a set of functions to manipulate the command list? This will make it easier to understand what functionality should be tested and maintained when making code changes.

```coffeescript
robot.registerCommand
  example: 'hubot: ping'
  description: 'Hubot will respond with PONG'
  module: 'hubot-ping'
robot.registeredCommands # returns an array of command examples
```

I'm not sure I completely like the functions as-is, but thinking in terms of functions instead of directly exposing data structures will make it a lot easier to maintain long term (you can see how hard changes are to directly exposed data structures in #917).

michaelansel:
cc #873 because `registerCommand` is starting to feel a lot like some of the ideas there

MrSaints:
:+1: for this idea. Coincidentally, I did the [exact same](https://github.com/ClaudeBot/hubot/commit/ccc3e7ca5a9f8e641a3ca33c45e73a891c7c5396) thing because I wanted an easier way to generate a list of configuration options available.

PR #1029, opened 2015-08-13
Remove unnecessary instance variable in Robot

michaelansel:
Introduced in 1e4ae576aa6fbcade3cd7204a6102e6ddbfca4c7 for no apparent reason. Given the amount of refactoring that has occurred to move each class into its own file, this should be completely unnecessary now.

### DIFF
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -1,6 +1,7 @@
 {inspect} = require 'util'
 async     = require 'async'
 
+Response = require './response'
 {TextMessage} = require './message'
 Middleware = require './middleware'
 
@@ -75,7 +76,7 @@ class Listener
         if cb?
           process.nextTick -> cb true
 
-      response = new @robot.Response(@robot, message, match)
+      response = new Response(@robot, message, match)
       middleware.execute(
         {listener: @, response: response}
         executeListener
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -46,7 +46,6 @@ class Robot
     @brain      = new Brain @
     @alias      = false
     @adapter    = null
-    @Response   = Response
     @commands   = []
     @listeners  = []
     @middleware =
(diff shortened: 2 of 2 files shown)


technicalpickles:
Off hand, I think we try to require the class just once, so `instanceof` correctly works. Like, if you `require 'response'` twice, the instanceof of the two responses is false. I think external adapters in particular had problems around this.

michaelansel:
That doesn't seem right to me... Due to the node module cache, if you `require 'response'` twice, you get a reference to the exact same object in memory. I'll do some testing to see if I can reproduce the issue though.

technicalpickles:
In https://github.com/github/hubot/pull/996#issuecomment-133707368 I suggested treating the current implementation as a feature. Feasibly, an adapter could implement it's own response class, and override `robot.Response`.

michaelansel:
Yikes, I really don't like that. Extending Hubot by modifying internal data structures terrifies me. It's a hack that would work, but is just going to turn into more weird interfaces we need to support. Anything we can do to standardize what is extensible and what isn't will make maintaining backwards-compatibility much easier.

michaelansel:
cc #1036 because that will really be the deciding factor

technicalpickles:
> Yikes, I really don't like that. Extending Hubot by modifying internal data structures terrifies me. It's a hack that would work, but is just going to turn into more weird interfaces we need to support. Anything we can do to standardize what is extensible and what isn't will make maintaining backwards-compatibility much easier.

Another way to think of it is that hubot has a factory for creating response objects. That makes it less scary to me. I think it makes sense that adapters have the first chance to customize it, since it's possible adapters would want richer output as talked about in https://github.com/github/hubot/issues/1036

technicalpickles:
I'd like to explore things like https://github.com/slackhq/hubot-slack/pull/234 which need `robot.Response` to be expose, so going to close.

Issue #823, opened 2014-11-25
Manually call a .respond listener


lukerollans:
I'm wondering if there's a best practice way of manually calling a `.respond` listener. In some instances, it would be great if I could manually call one and let params go through as default (eg, `msg.match[1]` for example will match nothing and fallback to default).

Normally, the action performed by the listener being called would be abstracted out but I'm just curious if anyone else has gone this way.

For example

``` coffeescript
module.exports = (robot) ->
  robot.respond /hello!/i, (msg) ->
      msg.send "Hello to you too, sir!"
      # Something here to call the below listener and have Hubot also wish the user a great day

  robot.respond /have a great day/i, (msg) ->
      msg.send "I wish upon you a fantastic day, as well!"
```

michaelansel:
As soon as #803 merges, I've got a [commit queued up](https://github.com/michaelansel/hubot/commits/issues/823) to add `robot.listenerById` which will retrieve a specific listener based on its ID (a new metadata field introduced in #803). You would then be able to execute the listener directly using the private API `call`.

In a unit test scenario:
```coffeescript
  describe 'listener "get-room-oncall"', ->
    it 'should match "who is oncall?"', (testDone) ->
      listener = @robot.listenerById 'get-room-oncall'

      # Stub out the real work; just testing the match
      listener.callback = sinon.stub() # private API

      testMessage = new TextMessage @user, "TestHubot: who is oncall?"

      listener.call testMessage, (result) -> # private API
        expect(result).to.be.ok
        testDone()
```

The relevant bit of script would be:
```coffeescript
module.exports = (robot) ->
  robot.respond /who is oncall?/, id: 'get-room-oncall', (response) ->
    # it does stuff
```

radius314:
i'm interested in being able to do this too, it seems kind of odd that I can't arbitrarily send a message back to hubot and chain event togethers.

michaelansel:
@radius314 What exactly are you trying to accomplish? While I hope to enable listener retrieval soon, I envision manual invocation as an edge case (`call` is a private API). Most reuse of logic should be done with functions, external modules, and [events](https://github.com/github/hubot/blob/master/docs/scripting.md#events).

michaelansel:
Opened #1031 to add `robot.listenerById`. When that merges, I'll close this.

spilio:
[...]

PR #1031, opened 2015-08-14
Add helper function to lookup Listeners

michaelansel:
Solves the last need required for a couple issues:
https://github.com/github/hubot/issues/823#issuecomment-73999270
https://github.com/github/hubot/pull/686#issuecomment-74805659

Allows retrieving the named Listener object for unit testing or advanced scripting.


technicalpickles:
We don't really talk about specs anywhere in the docs currently, so I'm not sure it's the best place to demonstrate `listenerById`. Maybe could include an example to invoke one listener for another?

(we should have docs on testing, just don't want to introduce here I don't think)

technicalpickles:
I think this should be good to merge. I would like to get more documentation around testing though, but it doesn't have to happen here.

timkinnane:
This looks very useful. Can we reopen please?

PR #1037, opened 2015-08-26
Tweak test case to fail better

michaelansel:
Test cases should fail fast and loud instead of by hitting a timeout


michaelansel:
@bhuga

bhuga:
:+1:

PR #1038, opened 2015-08-26
Receive middleware docs tweaks

bhuga:
Based on feedback from @michaelansel in https://github.com/github/hubot/pull/1019


michaelansel:
:+1:

Issue #1034, opened 2015-08-21
Proposal: Route Namespacing


gerred:
It would be nice for repetitive commands to have namespacing. For example, I'd like the following code:

``` coffeescript
robot.namespace /admin/, (admin) ->
  admin.reply /do a thing/, (res) ->
    # command replies to "hubot admin do a thing".

  admin.hear /let's have a party/, (res) ->
    # command replies to "admin let's have a party"
```

If people like it for inclusion into core, and we nail down an API, I can open a PR for it.

michaelansel:
Oooh, I really like this idea. cc #873, where I had some related ideas, but this feels super clean.

Namespaces could automatically add a prefix to the listener IDs as well.

gerred:
We may be able to combine #873 really nicely with this.

Perhaps some sort of context command, where all of the options you specified can be made in a declarative way.

For example, with my API proposal, you could do something like:

```coffeescript
robot.namespace /admin/, (admin) ->
  admin.config "ADMIN_TOKEN", required: true, description: "Here's some token", validation: (value) ->
  admin.namespace /hosts/, (hosts) ->
    hosts.config "AWS_ACCESS_KEY", required: true

    hosts.preConfig (env, done) -> # preconfiguration, perhaps to retrieve information for the subroutes
      setupSomething (resp) -> 
        env.region = resp.region
        done()

    hosts.respond /for tag (.*)=(.*)/, (msg) -> # respond/hear should wait for done callback if preConfig is defined
      label = res.match[1]
      value = res.match[2]
      aws.getHosts({region: msg.message.env.region, label: label, value: value}) # get context from higher namespaces
     .done (hosts) -> msg.reply(JSON.stringify hosts)
````

michaelansel:
I'll think on this more, but first pass: namespaces should probably be strings instead of regular expressions. Prevents weird problems like overlapping namespaces and allows for easy lookup of a namespace (imagine spreading listeners in the same namespace across multiple files or even multiple modules)

gerred:
:+1: on that.

bhuga:
I've got scripts that would benefit from this, but I think we can do even more. The usual pattern is:

 * Many commands start with a prefix.
 * All the commands use `msg.finish()`.
 * There's a final command that displays help if the prefix was used but no previous command matched.

This helps with the first of those three, but not the second two related issues. How would you feel about special semantics for the "matches a prefix but no other command" case?

technicalpickles:
[...]

PR #1048, opened 2015-09-14
File-scoped metadata

bhuga:
This PR adds `robot.listenerOptions metadataItem: true`. All listeners defined in a file after this call will extend this metadata rather than start from scratch. You can now do something like this:

```coffeescript
module.exports = (robot) ->
  robot.listenerOptions annoying: true
  robot.respond /annoy me/, (msg) ->
    # code to annoy someone. This listener will have annoying: true as its
    # metadata.

  robot.respond /unannoy me/, annoying: false, (msg) ->
    # Code to stop annoying someone. This lisenter will have annoying: false
    # as its metadata
```

@technicalpickles and I are running through a large existing codebase and found a few users for this. Some files have a lot of functions, and adding something like `deployment: true` to tag all commands related to deployment is pretty tedious and not very helpful. We're also enabling commands 1-by-1 for a new chat service after an audit, and adding `fooEnabled:` to every single command we have is getting ridiculous, and will result in a lot of diff churn when it's time to pull them out.

cc @michaelansel

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -638,6 +638,20 @@ module.exports = (robot) ->
     # code to stop annoying someone
 ```
 
+You can scope all listeners in a file with the `listenerOptions` function:
+
+```coffeescript
+module.exports = (robot) ->
+  robot.listenerOptions annoying: true
+  robot.respond /annoy me/, (msg) ->
+    # code to annoy someone. This listener will have annoying: true as its
+    # metadata.
+
+  robot.respond /unannoy me/, annoying: false, (msg) ->
+    # Code to stop annoying someone. This lisenter will have annoying: false
+    # as its metadata
+```
+
 These scoped identifiers allow you to externally specify new behaviors like:
 - authorization policy: "allow everyone in the `annoyers` group to execute `annoyance.*` commands"
 - rate limiting: "only allow executing `annoyance.start` once every 30 minutes"
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -27,6 +27,14 @@ class Listener
     if not @options.id?
       @options.id = null
 
+    if @robot.defaultOptions?
+      newOptions = {}
+      for key, val of @robot.defaultOptions
+        newOptions[key] = val
+      for key, val of @options
+        newOptions[key] = val
+      @options = newOptions
+
     if not @callback? or typeof @callback != 'function'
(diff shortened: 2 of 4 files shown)


technicalpickles:
The main issue I can spot would be scripts that asynchronously call `robot.respond`. Even if it's something like a short `setTimeout`, the robot. listenerOptions would have changed by the time it's invoked.

I was thinking about if the `modules.export = (robot) ->` pattern had an optional done argument to signal loading of the next script, but I can think of other ways that wouldn't be enough since it's intentionally defining listeners outside the context of loading a script. The main example I can think of is watching a resource, and registering new listeners dynamically. Not a pattern I've seen yet though.

michaelansel:
I think namespaces (#1034) is a muuuch better way of getting this behavior. In that model, you replace `robot` with a proxy object of some sort that contains all the shared config for the nested listeners. Anything that works by setting a global flag and depends on synchronicity scares the :hankey: out of me.

bhuga:
:cool: my ego is not at stake if we don't like this approach.

> In that model, you replace robot with a proxy object of some sort that contains all the shared config for the nested listeners. 

What I didn't like about #1034--and I know this is petty--is that it adds a layer of indentation to *every script ever*. Setting metadata or a namespace is data configuration, not an operation where we're waiting for a result.

It's also about namespacing listener regexes, and this is about metadata. But point taken: the proxy object approach is definitely more robust. We've already added `async`. I will see if I can't come up with gentler syntax for the proxy approach.

michaelansel:
> I will see if I can't come up with gentler syntax for the proxy approach.

Maybe something like this?

```coffeescript
module.exports = (robot) ->
  annoy = robot.createNamespace('annoy')

  annoy.listenerOptions =
    annoying: true

  annoy.respond /annoy me/, (msg) ->
    ...
```

At that point though, I'm immediately thinking it would be awesome to auto-create namespaces and provide them as a parameter alongside (or in place of) `robot`. This starts to get really close to the conversation in #873.

I haven't really had time recently to do any design work (and I don't see that changing for at least another couple weeks :sweat:), but I think there is an opportunity here to solve several different problems in a unified way that will set us up for some really nice future extensibility/compartmentalization.

bhuga:
Seems like namespaces will get us the right behavior better than this one would. When we land that one day, I'll poke around and try to make the syntax less indenty.

PR #1049, opened 2015-09-15
Fix link in heroku deployment guide

doque:


michaelansel:
:+1:

michaelansel:
@technicalpickles, while this does [fix the issue](https://hubot.github.com/docs/deploying/heroku/) on the docs site, it breaks in-repo links. Did we get to a point where everything should be working correctly in both?

technicalpickles:
We did. Maybe just linking to `/docs` would do it?

PR #1052, opened 2015-09-16
Only coerce ids that are all digits into integers

technicalpickles:
This lets you set a HUBOT_SHELL_USER_ID=U0123456, as if to pretend to
have a Slack user id.

Without this change, `parseInt` ends up returning undefined for strings like `U0123456`.

### DIFF
--- a/src/adapters/shell.coffee
+++ b/src/adapters/shell.coffee
@@ -42,7 +42,10 @@ class Shell extends Adapter
     @cli = cline()
 
     @cli.command '*', (input) =>
-      userId = parseInt(process.env.HUBOT_SHELL_USER_ID or '1')
+      userId = process.env.HUBOT_SHELL_USER_ID or '1'
+      if userId.match (/\A\d+\z/)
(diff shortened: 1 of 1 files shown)


michaelansel:
If we go this route, shouldn't we just make the IDs always strings instead of needing to deal with both data types? Probably doesn't matter much, but having mixed data types is always weird...

technicalpickles:
In practice, IDs always seem to be strings. I inspected our bot with the Campfire adapter, and the keys are strings of integers. Slack is strings already as I said. That leaves shell being the weird one, so I guess they should just always be strings.

technicalpickles:
The only annoyance with making them alwasys be strings in the shell adapter is that then there are brains out in the wild with integer keys, so that could be a bit of a surprise after an upgrade.

michaelansel:
My gut says that is okay as a breaking change since the shell adapter is never(?) used as a production adapter, but I do see the potential breakage...

PR #1054, opened 2015-09-22
Fix link to Deploying index

gregkare:
It's currently pointing to https://hubot.github.com/docs/deploying/index/ (which is a 404)


technicalpickles:
Fixed in https://github.com/github/hubot/pull/1116

Issue #923, opened 2015-04-20
impossible to use Hubot and receive secured webhook data


xrd:
I'm trying to use [secure webhooks](https://developer.github.com/webhooks/securing/) with Hubot.

I believe I have the correct algorithm for decrypting once I have the body content.

```
getSecureHash = ( body ) ->
        hmac = crypto.createHmac( 'sha1', _SECRET )
        hmac.setEncoding( 'hex' )
        hmac.write( body )
        hmac.end()
        hash = hmac.read()
        console.log "Hash: #{hash}"
        hash
```

However, if I understand correctly, I need to use the original complete body as sent into the HTTP request (something like `payload=%7B%22ref%22%3A%22refs%2Fheads%2F14...`) and verify that computing the HMAC hash matches the signature sent in the HTTP_X_HUB_SIGNATURE header.

However, Hubot does not (AFAICT) allow you to customize the middleware used by express. And, the bodyParser does not provide the raw body inside the request object. It does parse it as JSON and provide you with that as an inflated object inside the request object.

However, the JSON once parsed does not allow me to stringify it again and guarantee that it will match the original pre-parsed body. In other words, once the JSON is parsed by express, I lose any significant whitespace that was inside the URL encoded JSON string, and that whitespace will change the HMAC hash such that after a roundtrip I have no chance of matching HMAC signatures.

Is there a way to get the raw body content inside my router callback? If not, it seems like securing the webhook is not possible with Hubot.

sukima:
@xrd I can confirm this is the case. The bodyParser middleware gobbles up the data in the req stream. So the parsed result is available on `req.body` but the original raw body (the buffer data needed to compare the HMAC) is exhausted.

The only way to do this is to have a custom pass through middleware that will build the HMAC for you. Obviously Hubot core would need to be changed. Also since a custom webhook endpoint would have different secret keys there would have to be some mechanism to tell Hubot (and essentially the custom middleware) which key to use.

A double edged sword since `robot.router` is a reference to the express router allowing you to add your own middleware (`robot.router.use(…)`). However, the crux is that any use in a Hubot script is *after* the bodyParser middleware and the request stream would already have been exhausted.

And now I've rambled on till I've reached the conclusion that meta programming is the answer (add Hubot specific methods to the express.Router prototype). And at 2AM in the morning I'm way to tired for such thought experiments.

So yeah, If you thought GitHub's secure webhooks were a cool idea for your Hubot instance your pretty much :poop: out of luck at the moment.

[...]

PR #1056, opened 2015-09-25
Add a rawBody property to req object in custom HTTP listeners

brodul:
Hello,

I modified the express middleware that is used in custom HTTP listeners (routers) so that a hubot user can access the actual body of the request as a string object. 

The need for this is described in #923 . It's probably impossible to access the unprocessed body without monkey-patching hubot itself.

I need this for validating github webhooks that use HMAC.
I added an example in the documentation how to do this.

Accessing the body in express 3.x and testing middlewares is hard. After trying different approaches I decided to implement this solution: http://stackoverflow.com/a/26634372/807339 .

Mocha has a hard time detecting assert errors inside routers in the test cases. Therefore I had to rethrow them in the test context/namespace.

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -421,6 +421,37 @@ curl -d 'payload=%7B%22secret%22%3A%22C-TECH+Astronomy%22%7D' http://127.0.0.1:8
 
 All endpoint URLs should start with the literal string `/hubot` (regardless of what your robot's name is). This consistency makes it easier to set up webhooks (copy-pasteable URL) and guarantees that URLs are valid (not all bot names are URL-safe).
 
+### Get raw body string
+
+Hubot maps the raw body to the req.rawBody property. This is important,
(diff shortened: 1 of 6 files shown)


brodul:
The tests fail for node.js 0.8 but it seems as a travis env problem.
http://stackoverflow.com/questions/31025048/npm-doesnt-work-get-always-this-error-error-cannot-find-module-are-we-ther

https://travis-ci.org/github/hubot/jobs/82107102

brodul:
HI,

here is some :cake: :smile:

brodul:
@bkeepers If you are considering reviewing this PR, I can fix it.

timkinnane:
@brodul, would updating express to v4 help you with this? Just linking to #1379 in case these are related. I know we had a similar issue with web-hook authentication that we hacked around but I think express v4 would have helped.

PR #1050, opened 2015-09-15
update travis.yml

scboucher:
Added node v4 and node v0.12, removed  node v0.11

PR #1068, opened 2015-10-13
Add node up to v4.2 to travis. Add intellij stuff to gitignore

scboucher:
See issue #1050

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 language: node_js
-node_js:
-  - "0.11"
-  - "0.10"
+
 matrix:
   include:
     - node_js: "0.8"
+      before_install: npm  update -g  npm
       # old versions of NPM don't include support for `^` dependency prefixes in package.json
-      before_install: npm update -g npm
+    - node_js: "0.10"
+    - node_js: "0.12"
+    - node_js: "4.0"
+    - node_js: "4.1"
+    - node_js: "4.2"
 notifications:
   email: false
+
+sudo: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,11 +1,16 @@
...
 matrix:
   include:
     - node_js: "0.8"
+      before_install: npm  update -g  npm
       # old versions of NPM don't include support for `^` dependency prefixes in package.json
-      before_install: npm update -g npm
+    - node_js: "0.10"
+    - node_js: "0.12"
+    - node_js: "4.0"
+    - node_js: "4.1"
```
technicalpickles: Are odd releases still considered 'unstable'? If so, I don't think we need to include that here.

scboucher:
Maybe we could only include v4.2.* since it is the LTS version. But there is no more unstable release. Instead odd version are not LTS.
https://nodesource.com/assets/blog/essential-steps-lts/nodejs_lts_schedule.png
https://nodejs.org/en/blog/release/v4.2.1/
https://nodesource.com/blog/essential-steps-long-term-support-for-nodejs

scboucher:
I thought about it, maybe the best solution is to only target node v4. I don't believe there will be any major change between v4.2.X and v5.
Then the question is should Hubot support only lts version or all of them.

I'll try to investigate the v0.8 problem it seems to be related to npm. But for now I think it is the best to allow failure on this version.

scboucher:
Removed node v0.8 cc #1070

michaelansel:
I'm a fan. I don't normally like adding IDE files to the git ignore (should be using a local global ignore for that), but I'll let it go as long as @technicalpickles doesn't object.

michaelansel:
:+1:

Issue #919, opened 2015-04-14
Reconnecting causes double load of scripts


michaelansel:
From [a comment on #71](https://github.com/github/hubot/issues/71#issuecomment-57559761):

Scripts appear to double load (resulting in double responses for everything) whenever connection is lost and reestablished.

My original hypothesis:

> The double load on reconnect is likely caused by the 'connected' event getting emitted again. The robot listens for 'connected' and loads all scripts at that point (assuming it will only get fired once). This would be adapter dependent (adapter decides when to emit 'connected').

However, it looks like the HipChat adapter handles this case already: https://github.com/hipchat/hubot-hipchat/blob/master/src/hipchat.coffee#L115

Not sure what is going on here...

michaelansel:
@lukaszraczylo What adapter are you using when you see the issue?

technicalpickles:
Other thoughts:

* could it be related to a script reloading script? ie reload.coffee in hubot-scripts
* could it be related to compiling coffeescript to javascript and both of them are loaded? I seem to recall a bug in that area in the past.

lukaszraczylo:
@michaelansel hipchat adapter as well. Sorry for delay.

apottere:
We've been seeing double responses intermittently for a while now, and I think this is the issue.  Bot will be up for a few days, and one morning will just start giving duplicated responses.  A restart fixes the issue.

apottere:
Not sure if this belongs here, but these are the logs right before duplicated messages start appearing:

````
[Fri Apr 17 2015 15:17:35 GMT+0000 (UTC)] INFO Disconnected from hipchat.com
[Fri Apr 17 2015 15:17:35 GMT+0000 (UTC)] INFO Waiting 13s and then retrying...
[Fri Apr 17 2015 15:17:35 GMT+0000 (UTC)] ERROR null, null, { name: 'stream:error',
  parent: null,
  attrs: { 'xmlns:stream': 'http://etherx.jabber.org/streams' },
  children:
   [ { name: 'policy-violation',
       parent: [Circular],
       attrs: { xmlns: 'urn:ietf:params:xml:ns:xmpp-streams' },
       children: [] },
     { name: 'text',
       parent: [Circular],
       attrs: { xmlns: 'urn:ietf:params:xml:ns:xmpp-streams' },
       children: [ 'Stanza too big.' ] } ] }
[Fri Apr 17 2015 15:17:35 GMT+0000 (UTC)] INFO Connection was closed
[...]

PR #1069, opened 2015-10-20
Only loadScripts once

technicalpickles:
This should fix https://github.com/slackhq/hubot-slack/issues/210 over in hubot-slack. I'm not sure if there's any other implications of this, like if other adapters have workarounds for this or not.

PR #1073, opened 2015-10-27
Only load scripts the first time the adapter connects

Floby:
This should fix issue #919 where hubot reloads scripts
each time the adapter emits `'connected'`

Because some adapters handle retries for lost connections, they
may emit such an event many times. However hubot should not load
scripts more than once.

No tests added, `bin/hubot` is hardly testable and the modification
does not break any existing test

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -132,6 +132,6 @@ else
     console.log "OK"
     process.exit 0
 
-  robot.adapter.on 'connected', loadScripts
+  robot.adapter.once 'connected', loadScripts
 
   robot.run()


apottere:
+1

Floby:
the automatic checks say I didn't agree to the Github CLA but I did :(

technicalpickles:
This is a duplicate of https://github.com/github/hubot/pull/1069

Floby:
yes.

Floby:
but mine passes travis somhow :/

technicalpickles:
@Floby we dropped node 0.8 for travis between my PR and yours.

PR #1077, opened 2015-11-03
Add node v5 to tests.

scboucher:
I added stable to travis as  it is currently pointing to node v5. 
I added dev dependencies to support npm v3 (sinon and chai were require but not found.) I also updated the libs. There is one test pending.


bhuga:
> I added stable to travis as it is currently pointing to node v5.

Can we point to a specific version?

scboucher:
Yes we can, also I saw that the nvm stable alias is deprecated

PR #1090, opened 2015-11-20
[+] Included Mattermost Hubot adapter

heitorlessa:
Include [Mattermost](mattermost.org) Hubot adapter in the list


technicalpickles:
We are trying to keep this file alphabetical actually. Would you mind updating to reflect that?

technicalpickles:
Handling this in https://github.com/github/hubot/pull/1120

PR #1097, opened 2015-12-12
Add Skyweb to a list of adapters

EllisV:
Hello,

I would like to add a [Skyweb Adapter](https://github.com/EllisV/hubot-skyweb) which uses Skyweb library (communicates with Skype via HTTP) to a list of adapters.


technicalpickles:
Thanks!

Issue #1093, opened 2015-11-29
Rename Hubot at runtime


pchaigno:
Is there a way to rename Hubot at runtime (using a script)?
I tried the following code with no effect.

``` coffee
robot.name = 'hubot_'
```

How should I proceed if I want to add this functionality?

bhuga:
This will be pretty tough. `robot.respond` invocations resolve hubot's name at load time and create a regex from it. Each `respond` listener's regex would need to be updated, and it's not particularly simple to identify `respond` vs `hear`.

The guts of this behavior are in in `respond` and `responsePattern` in `robot.coffee` if you want to dig deeper.

pchaigno:
Wouldn't reloading the scripts achieve the same result?

michaelansel:
@pchaigno yes, but reloading is non-trivial and currently unsupported cc #715

PR #1104, opened 2016-01-02
Rename Hubot at runtime

pchaigno:
First implementation for #1093, this pull request adds a `changeName` to Robot to change its name at runtime. It updates the name and the `respond` listeners.
It replaces the beginning of listeners' regular expressions, which contains the name of the robot, instead of reconstructing them.

I'm not a big fan of this piece of code because it shares a lot with `respondPattern` and, it depends heavily on the `respond` regex.
An alternate solution is to reconstruct regular expressions using `respondPattern`. This strategy would, however, require to save the original regex (the message part that follows the robot's name/alias) in the TextListener object.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -148,6 +148,47 @@ class Robot
 
     newRegex
 
+  # Public: Updates the robot name and its listeners accordingly.
+  #
+  # newName - A String of the new robot name.
+  #
+  # Returns nothing.
+  changeName: (newName) ->
+    # Escapes names:
+    alias = @alias.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&')
+    oldNameRegex = @name.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&')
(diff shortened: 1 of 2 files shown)


technicalpickles:
Thanks for taking a stab at this implementation, and :sparkles: for including tests!

> An alternate solution is to reconstruct regular expressions using respondPattern. This strategy would, however, require to save the original regex (the message part that follows the robot's name/alias) in the TextListener object.

I share your concerns, and think this would be a better approach :+1: 

If people change the name, do you think they'd want to change the alias too?

Another thing I would think about is documentation. This should be included in `scripting.md` if it's to be used by scripts. That'd be a good place to talk about when you'd want to use this (or when you wouldn't).

I'd also wonder if there's any caveats to supporting this. Imagine I'm talking to hubot in one room, and someone changes the name somewhere else. I'd expect hubot to still be respond, but it'd start silently ignoring me until I found the new name.

pchaigno:
> I share your concerns, and think this would be a better approach :+1: 

I implemented it at #1115. I'll close this one.
I also tried to answer your questions/concerns in the new PR.

PR #1106, opened 2016-01-05
Use env var for setting limit and parameterLimit to express's urlenconder

### LINKED ISSUE #1105: Modify the default parameterLimit value in robot's express
Right now I use Hubot with `Chargify` like webhook and I am getting some `413 Request entity too large` with some events.
Digging we found that here: https://github.com/github/hubot/blob/master/src/robot.coffee#L430 , `robot` initialize the `express` object and call `app.use express.urlencoded()` without options. 

How can we modify the `paramLimit` for not getting `413` often: https://github.com/expressjs/body-parser#bodyparserurlencodedoptions ?

gotchahn:
For avoiding `413 Request Entity too large` in some responses.
fixes #1105

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -429,7 +429,7 @@ class Robot
     app.use express.query()
 
     app.use express.json()
-    app.use express.urlencoded({limit: limit, parameterLimit: paramLimit})
+    app.use express.urlencoded(limit: limit, parameterLimit: paramLimit)
     # replacement for deprecated express.multipart/connect.multipart
     # limit to 100mb, as per the old behavior
     app.use multipart(maxFilesSize: 100 * 1024 * 1024)
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -414,7 +414,7 @@ class Robot
     port       = process.env.EXPRESS_PORT or process.env.PORT or 8080
     address    = process.env.EXPRESS_BIND_ADDRESS or process.env.BIND_ADDRESS or '0.0.0.0'
     limit      = process.env.EXPRESS_LIMIT or '100kb'
-    paramLimit = process.env.EXPRESS_PARAMETER_LIMIT or 1000
+    paramLimit = parseInt(process.env.EXPRESS_PARAMETER_LIMIT) or 1000
 
     express = require 'express'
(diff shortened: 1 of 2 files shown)


src/robot.coffee
```diff
@@ -408,11 +408,13 @@ class Robot
...
-    stat    = process.env.EXPRESS_STATIC
-    port    = process.env.EXPRESS_PORT or process.env.PORT or 8080
-    address = process.env.EXPRESS_BIND_ADDRESS or process.env.BIND_ADDRESS or '0.0.0.0'
+    user       = process.env.EXPRESS_USER
+    pass       = process.env.EXPRESS_PASSWORD
+    stat       = process.env.EXPRESS_STATIC
+    port       = process.env.EXPRESS_PORT or process.env.PORT or 8080
+    address    = process.env.EXPRESS_BIND_ADDRESS or process.env.BIND_ADDRESS or '0.0.0.0'
+    limit      = process.env.EXPRESS_LIMIT or '100kb'
+    paramLimit = process.env.EXPRESS_PARAMETER_LIMIT or 1000
```
parkr: Does this have to be run through `parseInt`?

gotchahn: I think you are right. If the env `EXPRESS_PARAMETER_LIMIT` is not a number, `NaN` or `undefined` will return just `NaN` and that `or` 1000 will give us `1000`. Good catch! :+1:

technicalpickles: TIL `NaN` is falsey.

src/robot.coffee
```diff
@@ -427,7 +429,7 @@ class Robot
     app.use express.query()
 
     app.use express.json()
-    app.use express.urlencoded()
+    app.use express.urlencoded({limit: limit, parameterLimit: paramLimit})
```
technicalpickles: :coffee:ism, but you don't need `{}` here.

technicalpickles:
I'm okay adding this, but would you mind updating the documentation to talk about these new environment variables? https://github.com/github/hubot/blob/master/docs/scripting.md#http-listener makes sense.

gotchahn:
Sound good!

gotchahn:
UPDATED :boom:

PR #1109, opened 2016-01-08
Determine adapterPath in robot.coffee, rather than bin/hubot

technicalpickles:
Previously, adapterPath was determined in bin/hubot, and passed to index.coffee,
 and eventually passed ot the constructor in robot.coffee. This is used by Robot
to determine where to load builtin adapters from.

It also makes it hard to progmatically construct a robot without running bin/hubot,
since you'd need to find a way to figure out the path that the hubot module is installed.

This changes it so Robot will determine the adapterPath since it's relative to where
it live. The constructor still takes it as an option so it remains backward compatible
 to anything that might be specifying it.

This is work towards https://github.com/github/hubot/issues/858

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -34,13 +34,15 @@ class Robot
   # Robots receive messages from a chat source (Campfire, irc, etc), and
   # dispatch them to matching listeners.
   #
-  # adapterPath - A String of the path to local adapters.
+  # adapterPath -  A String of the path to built-in adapters (defaults to src/adapters)
   # adapter     - A String of the adapter name.
   # httpd       - A Boolean whether to enable the HTTP daemon.
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
   constructor: (adapterPath, adapter, httpd, name = 'Hubot', alias = false) ->
+    @adapterPath ?= Path.join __dirname, "adapters"
+
     @name       = name
     @events     = new EventEmitter
     @brain      = new Brain @
@@ -62,7 +64,7 @@ class Robot
     else
       @setupNullRouter()
 
-    @loadAdapter adapterPath, adapter
+    @loadAdapter adapter
 
     @adapterName   = adapter
     @errorHandlers = []
@@ -452,12 +454,12 @@ class Robot
   # adapter - A String of the adapter name to use.
   #
   # Returns nothing.
-  loadAdapter: (path, adapter) ->
+  loadAdapter: (adapter) ->
     @logger.debug "Loading adapter #{adapter}"
 
     try
       path = if adapter in HUBOT_DEFAULT_ADAPTERS
-        "#{path}/#{adapter}"
+        "#{@adapterPath}/#{adapter}"
       else
         "hubot-#{adapter}"
 
--- a/bin/hubot
+++ b/bin/hubot
@@ -82,9 +82,7 @@ if Options.create
   process.exit 1
 
 else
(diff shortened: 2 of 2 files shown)


technicalpickles:
This allows you to have a file like `index.coffee`:

```coffeescript
Hubot = require 'hubot'

robot = new Hubot.Robot(null, 'shell', false, 'Hubot', '/')

robot.respond /ping/i, (res) ->
  res.send 'PONG'

require('hubot-shipit')(robot)

robot.run()
```

And to it with `coffee index.coffee`.

technicalpickles:
I was able to take that example a bit further:

```
Hubot = require 'hubot'
Path  = require 'path'
​
robot = new Hubot.Robot(null, 'shell', false, 'Hubot', '/')
​
# wait for adapter to be connected, since it might change things at runtime,
# like robot name
robot.adapter.on 'connected', ->
​
  # require external hubot scripts
  require('hubot-help')(robot)
  require('hubot-shipit')(robot)
​
  # load scripts from a local scripts directory
  # NOTE you need to either specify NODE_PATH=. or Path.resolve a local directory for this to work
  robot.load Path.resolve 'scripts'
​
  # or do things them inline
  robot.respond /ping/i, (res) ->
    res.send 'PONG'
​
# then run hubot
robot.run()
```

src/robot.coffee
```diff
@@ -34,13 +34,15 @@ class Robot
...
   #
-  # adapterPath - A String of the path to local adapters.
+  # adapterPath -  A String of the path to built-in adapters (defaults to src/adapters)
   # adapter     - A String of the adapter name.
   # httpd       - A Boolean whether to enable the HTTP daemon.
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
   constructor: (adapterPath, adapter, httpd, name = 'Hubot', alias = false) ->
+    @adapterPath ?= Path.join __dirname, "adapters"
```
michaelansel: Given that the list of adapters is hard coded as a constant (used on L461), it seems to me the adapter path should be hard coded as well... I guess we're maintaining backwards compatibility, but I can't ever see a situation when you would actually want to specify a custom `adapterPath`.

gruberjuliannama:
new Hubot.Robot(null, 'shell', false, 'Hubot', '/')
​
# wait for adapter to be connected, since it might change things at runtime,
# like robot name
robot.adapter.on 'connected', ->
​
  # require external hubot scripts
  require('hubot-help')(robot)
  require('hubot-shipit')(robot)
​
  # load scripts from a local scripts directory
  # NOTE you need to either specify NODE_PATH=. or Path.resolve a local directory for this to work
  robot.load Path.resolve 'scripts'
​
  # or do things them inline
  robot.respond /ping/i, (res) ->
    res.send 'PONG'
​
# then run hubot

gruberjuliannama:
= require 'hubot'

robot = new Hubot.Robot(null, 'shell', false, 'Hubot', '/')

robot.respond /ping/i, (res) ->
  res.send 'PONG'

require('hubot-shipit')(robot

gruberjuliannama:
adapter     - A String of the adapter name.
   # httpd       - A Boolean whether to enable the HTTP daemon.
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
   constructor: (adapterPath, adapter, httpd,

gruberjuliannama:
I guess we're maintaining backwards compatibility, but I can't ever see a situation when you would actually want to.defaults to Hubot.

PR #1110, opened 2016-01-08
Programmatically configured hubot

technicalpickles:
Once I got https://github.com/github/hubot/pull/1109 working in a single file, started to pick at some of the API that didn't feel great:
- `Robot` constructor has positional arguments, which means they don't make much sense when you are looking at `new Hubot.Robot(null, 'shell', false, 'Hubot', '/')`
- `adapter` is loaded at Robot construction time. the `loadAdapter` is just a wrapper around loading a builtin adapter vs a npm module
- `loadAdapter` expects an adapter's exports to define a `use` method, and in reality, that seems to only ever return a new instance of the adapter that takes a robot. Why not construct this directly?
- loading an npm script package looks weird. we require it only to call it with the robot. this is the problem of the script API exporting a function though

What this does so far:
- constructor can now be called without any args
- update builtin adapter to export the adapter class, and make those accessible from top-level `require 'hubot'`
- don't load adapter if one wasn't given (in the case of constructor without any arg)
- add `loadScriptPackage` helper, which reads a little better but doesn't do much

Some open-ended questions / problems:
- making sure to preserve backwards compatibility
- handling environment specific configuration. should you hard-code name, alias, etc? environment variable? assign from an environment variable? should there be like a `config/environments.coffee` like Rails has, and `config/environments/development.coffee` for dev, `config/environments/production.coffee` for production, etc?
- does this make it easier to test? harder?
- does `bin/hubot` need to exist? does it need to support a programatic launch, or do you only use progmatic launch OR bin/hubot?

cc https://github.com/github/hubot/issues/858

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -62,7 +62,7 @@ class Robot
     else
       @setupNullRouter()
 
-    @loadAdapter adapter
+    @loadAdapter adapter if adapter?
 
     @adapterName   = adapter
     @errorHandlers = []
@@ -568,6 +568,9 @@ class Robot
   #
   # Returns nothing.
   run: ->
+    unless @adapter
+      throw new Error("no adapter present. did you forget to specify one or not call 'loadApdapter'?")
+
     @emit "running"
     @adapter.run()
 
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -448,7 +448,6 @@ class Robot
 
   # Load the adapter Hubot is going to use.
   #
-  # path    - A String of the path to adapter if local.
   # adapter - A String of the adapter name to use.
   #
   # Returns nothing.
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -34,13 +34,15 @@ class Robot
   # Robots receive messages from a chat source (Campfire, irc, etc), and
   # dispatch them to matching listeners.
   #
-  # adapterPath - A String of the path to local adapters.
+  # adapterPath -  A String of the path to built-in adapters (defaults to src/adapters)
   # adapter     - A String of the adapter name.
   # httpd       - A Boolean whether to enable the HTTP daemon.
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
   constructor: (adapterPath, adapter, httpd, name = 'Hubot', alias = false) ->
+    @adapterPath ?= Path.join __dirname, "adapters"
+
     @name       = name
     @events     = new EventEmitter
     @brain      = new Brain @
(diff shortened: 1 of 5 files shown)


technicalpickles:
Here's an example hubot:

```coffeescript
Hubot   = require 'hubot'
Log     = require 'log'
Path    = require 'path'
Express = require 'express'
BodyParser = require('body-parser')

robot = new Hubot.Robot
robot.logger = new Log('info')
robot.adapter = new Hubot.BuiltinAdapters.Shell(robot)
robot.alias = '/'

app = Express()
app.use (req, res, next) =>
  res.setHeader "X-Powered-By", "hubot/#{@name}"
  next()

app.use Express.query()

app.use BodyParser.json()
app.use BodyParser.urlencoded()

robot.router = app

# wait for adapter to be connected, since it might change things at runtime,
# like robot name
robot.adapter.on 'connected', ->

  # require external hubot scripts
  robot.loadScriptPackage require('hubot-help')
  robot.loadScriptPackage require('hubot-shipit')

  # load scripts from a local scripts directory
  # NOTE you need to either specify NODE_PATH=. or Path.resolve a local directory for this to work
  robot.load Path.resolve 'scripts'

  # or do things them inline
  robot.respond /ping/i, (res) ->
    res.send 'PONG'

  robot.router.get '/', (req, res) ->
    res.send('Hello World!')

  robot.router.listen(8081)

# then run hubot
robot.run()

```

michaelansel:
cc #1134 ; we should make sure the two big refactoring projects don't conflict (I think they're okay right now)

src/robot.coffee
```diff
@@ -568,6 +583,9 @@ class Robot
   #
   # Returns nothing.
   run: ->
+    unless @adapter
+      throw new Error("no adapter present. did you forget to specify one or not call 'loadApdapter'?")
```
michaelansel: typo :grin:

src/robot.coffee
```diff
@@ -376,17 +376,33 @@ class Robot
...
+  # Examples:
+  #   robot.loadScriptPackage require('hubot-help')
+  #   robot.loadScriptPackage 'hubot-help'
+  #
+  # Note:
+  #   Specifying package as string may fail depending on your NODE_PATH settings.
+  #   This can be fixed by requiring the package, and passing that to this function.
+  #
+  # Returns nothing
+  loadScriptPackage: (pkg, scripts) ->
```
michaelansel: Shouldn't we also be using this for `loadHubotScripts` and potentially for `loadFile`? Consolidating all calls to `require` would make the `new` refactor a lot easier as well as provide a consistent interface for programatic loading of any type of script.

Issue #1102, opened 2015-12-19
hubot shell adapter output hard to read on light terminal


glensc:
hubot shell adapter output hard to read on light terminal

![screenshot at 2015-12-19 02 06 30](https://cloud.githubusercontent.com/assets/199095/11909865/3bbc01de-a5f5-11e5-8c61-2ed4c4a9df0b.png)

that bold green is too light and small contrast

technicalpickles:
How does bold text look for you? Right now the output is bold green, but we can remove green. Relevant code is https://github.com/github/hubot/blob/master/src/adapters/shell.coffee#L20

glensc:
bold text should be ok, as it will be just black

PR #1111, opened 2016-01-08
[shell] don't color hubot responses green, to be more visible on light backgrounds

technicalpickles:
This is a possible fix for https://github.com/github/hubot/issues/1102

cc @glensc


glensc:
:+1:

technicalpickles:
I have a feeling I'll want to revisit this because the bold text doesn't really stand out without color. Maybe have environment variable like HUBOT_SHELL_OUTPUT_COLOR?

PR #1115, opened 2016-01-09
Rename Hubot at runtime using respondPattern

pchaigno:
Second implementation for #1093 (first was #1104); this pull request adds a `changeName` method to Robot to change its name at runtime. It updates the name and the `respond` listeners.
It reconstructs listeners using `respondPattern`. The original regex are saved in the `RespondListener` objects.

Should `respond` also notify the `@adapter` of the name change? (I suppose the server, too, needs to change the bot's name.)
<br>

My use case for this has to do with IRC. Under certain conditions, Hubot can be assigned an alternate nickname when connecting to the server. I'd like to be able to reassign the primary name when it becomes possible, without restarting Hubot.
I imagine this could prove useful in other situations.
<br>

From the first pull request (#1104):

> If people change the name, do you think they'd want to change the alias too?

I can't think of a case where the alias would need to be changed regularly, but maybe that should be possible, at least to be consistent.

> I'd also wonder if there's any caveats to supporting this. Imagine I'm talking to hubot in one room, and someone changes the name somewhere else. I'd expect hubot to still be respond, but it'd start silently ignoring me until I found the new name.

Maybe Hubot could announce his new name?
<br>

Note: I haven't updated `scripting.md` yet.

### DIFF
--- a/src/listener.coffee
+++ b/src/listener.coffee
@@ -104,7 +104,21 @@ class TextListener extends Listener
         message.match @regex
     super @robot, @matcher, @options, @callback
 
+class RespondListener extends TextListener
+  # RespondListeners receive every message from the chat source and decide if they
+  # want to act on it.
+  #
+  # robot    - A Robot instance.
+  # regex    - A RegExp for the message part that follows the robot's name/alias.
+  # options  - An Object of additional parameters keyed on extension name
+  #            (optional).
+  # callback - A Function that is triggered if the incoming message matches.
+  constructor: (@robot, @originalRegex, @options, @callback) ->
+    @regex = @robot.respondPattern(@originalRegex)
+    super @robot, @regex, @options, @callback
+
 module.exports = {
   Listener
   TextListener
+  RespondListener
 }
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -8,7 +8,7 @@ async          = require 'async'
 User = require './user'
 Brain = require './brain'
 Response = require './response'
-{Listener,TextListener} = require './listener'
+{Listener,TextListener,RespondListener} = require './listener'
 {EnterMessage,LeaveMessage,TopicMessage,CatchAllMessage} = require './message'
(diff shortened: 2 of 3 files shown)


technicalpickles:
> My use case for this has to do with IRC. Under certain conditions, Hubot can be assigned an alternate nickname when connecting to the server. I'd like to be able to reassign the primary name when it becomes possible, without restarting Hubot.

Thanks for sharing the use case, that helps a lot.

I was doing some commit history spelunking Friday to figure out we have code like https://github.com/github/hubot/blob/master/bin/hubot#L135 to only load scripts AFTER the adapter has connected. git blame brings us back to https://github.com/github/hubot/pull/220 which sites this as a primary reason:

> Adapters don't have a chance to customize the robot before scripts start loading.
>This can cause problems if the adapter determines the robot's name.

If that's the case, I suspect the IRC adapter could get a name from the server and update the `robot` before emitting the `connected` event.

technicalpickles:
After I hit comment, it occurred to me even if an irc hubot got a different name at startup (ie `hubot_` or whatever convention), it'd probably want to change it's name to it's proper name once it's available. That means this rename at runtime could still be useful.

pchaigno:
> If that's the case, I suspect the IRC adapter could get a name from the server and update the robot before emitting the connected event.

Thanks for the explanation :+1: 
I'll look into it and will open a pull request in the repository of the IRC adapter ;)

> After I hit comment, it occurred to me even if an irc hubot got a different name at startup (ie hubot_ or whatever convention), it'd probably want to change it's name to it's proper name once it's available. That means this rename at runtime could still be useful.

Actually, that was exactly my use case before I started investigating. I hadn't noticed that Hubot internal name was different from the server's.

michaelansel:
doh, I forgot about documentation! 😭 

When I mentioned edge cases, I was referring to bugs that arise as a result of changing the name.

PR #1123, opened 2016-01-24
Update azure.md

xt0rted:
Fixed the quotes in the deploy script step, and the name of the azure blob storage brain as per [hubot-azure-scripts](https://github.com/bfcamara/hubot-azure-scripts/blob/2105b4cb3dc331157e19c90e6dabf2247e6553a6/brain/storage-blob-brain.coffee)

### DIFF
--- a/docs/deploying/azure.md
+++ b/docs/deploying/azure.md
@@ -35,7 +35,7 @@ First, run the follow command to add `deploy.cmd` to your hubot directory. This 
 Then, edit this file and look for the sections that give you steps 1, 2 and 3. You're going to add a 4th step:
 
     :: 4. Create Hubot file with a coffee extension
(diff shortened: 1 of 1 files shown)


xt0rted:
While setting this up on azure I also found that there was an additional step of updating the node version in `iisnode.yml` to match that of the `WEBSITE_NODE_DEFAULT_VERSION` environment variable. The value in `iisnode.yml` was `0.10.0` while azure was set to use `4.2.4`. Hubot wouldn't run until these versions matched.

gep13:
@xt0rted did you have any issues around the following:

```
Application has thrown an uncaught exception and is terminated:
Error: Cannot find module 'coffee-script/register'
```

Did I miss something when setting this up?

technicalpickles:
It seems the hubot-azure-scripts package changed the path. I can merge this, but if you see any other problems deploying on Azure @gep13  or @xt0rted, can you please open new issues/PRs?

gep13:
@technicalpickles we were actually able to get this working.  And the steps documented here: https://github.com/Odonno/gitter-bot-how-to#host-on-azure

Thanks!

PR #1128, opened 2016-02-04
Add support for Babel 6 export default

kossnocorp:
Hey, I'm the author of [babel-hubot](https://github.com/kossnocorp/babel-hubot), minimalistic Hubot wrapper that allows to use Babel to compile scripts written in JS.

Prior to Babel 6, integration was super simple:

``` diff
--- a/bin/hubot
+++ b/bin/hubot
@@ -3,4 +3,4 @@
 npm install
 export PATH="node_modules/.bin:node_modules/hubot/node_modules/.bin:$PATH"

-exec node_modules/.bin/hubot "$@"
+exec node_modules/.bin/babel-hubot "$@"
```

Unfortunately in 6th version, Babel [changed the way how `export default` is handled](http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default), so now:

``` js
export default function() {}
```

... is not equal to:

``` js
module.exports = function() {}
```

... but to:

``` js
module.exports = {default: function() {}}
```

To make it work properly I had to change Hubot itself and check if `script.default` is a function and then use it instead of `script`.

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -354,6 +354,10 @@ class Robot
       try
         script = require(full)
 
+        # Support Babel 6's `export default function`
+        if typeof script.default is 'function'
+          script = script.default
+
         if typeof script is 'function'
           script @
           @parseHelp Path.join(path, file)
--- a/test/robot_test.coffee
+++ b/test/robot_test.coffee
@@ -341,6 +341,22 @@ describe 'Robot', ->
           @robot.loadFile('./scripts', 'test-script.coffee')
           expect(@robot.parseHelp).to.have.been.calledWith('scripts/test-script.coffee')
 
+      describe "script generated using Babel 6's export default", ->
+        beforeEach ->
+          module = require 'module'
+
+          @script = default: sinon.spy((robot) ->)
+          @sandbox.stub(module, '_load').returns(@script)
+          @sandbox.stub @robot, 'parseHelp'
+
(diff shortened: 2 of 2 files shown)


kossnocorp:
Any chance to get it merged?

conf:
Good job! :+1:

michaelansel:
This is an change/extension of the existing script interface. Currently, the sole and standard way of providing a script to Hubot is by exporting a function. If we need a more flexible script interface, I would much prefer to focus our energy on making that generic through programmatic loading (#1134).

Can you give me an idea of how impactful this is to bot owners to not upgrade to Babel 6 as compared to things like upgrading the version of coffeescript? For the coffeescript upgrade, there were new features that were really nice, but you could continue using the old version without a problem.

timkinnane:
I unknowingly duplicated this issue with #1374. Would be nice to get some action on allowing es6 scripts. @kossnocorp Am I right that would your solution would not effect any existing interfaces, but just allows the es6 default export format as well? @michaelansel If this solution provides some future-proofing to the script loading interface, without depreciating the prior format, why not allow it in the mean time, because coming up with a new more considered interface could take a long time.

@kossnocorp - have you tried using the babel `add-module-exports` plugin? https://github.com/59naga/babel-plugin-add-module-exports - as a stop-gap, it should convert your default exports into standard es5 module exports.

PR #1133, opened 2016-02-17
Correct documentation & variable names of functions that take envelopes, not users

technicalpickles:
Awhile back in hubot's history, adapters were changed to take an
'envelope' instead of a user or room object. The adapter class was
updated but not the robot functions that delegate to the adapter.

I noticed this when someone was adding a script using it, and I was checking the usage.


michaelansel:
:+1: yay!

PR #1134, opened 2016-02-19
Hubot, a Chatbot in ~Three~ Four Parts.

ajacksified:
Refactor `robot` into four files:
- A very basic robot class which manages the running of middleware, the event emitter, and other core functionality;
- One class subclasses the base robot class and provides autoloading scripts, "help" documentation, and auto-starting an express server;
- express.coffee, which contains functionality to create the web server
- scopedHTTPClient.coffee, which contains functionality for `robot.http`

**There should be no user-facing changes to how hubot works**. bin/hubot works as normal; http servers, aliases, etc use the same defaults as before. (It does expose a `setupRouter` function, but that functionality already existed by overwriting `@router` anyways.)

The purpose of this is so that one can use the "basebot" for all of hubot's functionality, without having to also require autoloading of files. For example, one may want to precompile hubot + scripts into a single, dead-code-eliminated javascript file for lower memory usage and startup time (which is my particular usecase- precompile the coffeescript.) Or, maybe you want a client-side hubot loaded because you're a mad genius doing crazy things. _many possibilities._

We could also call basebot "batteries" or something clever.

<3 hubot is the best

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -1,6 +1,7 @@
 User                                                                 = require './src/user'
 Brain                                                                = require './src/brain'
 Robot                                                                = require './src/robot'
+Basebot                                                              = require './src/basebot'
 Adapter                                                              = require './src/adapter'
 Response                                                             = require './src/response'
 {Listener,TextListener}                                              = require './src/listener'
@@ -9,6 +10,7 @@ Response                                                             = require '
 module.exports = {
   User
   Brain
+  Basebot
   Robot
   Adapter
   Response
--- a/src/basebot.coffee
+++ b/src/basebot.coffee
@@ -1,5 +1,4 @@
 Log            = require 'log'
-HttpClient     = require 'scoped-http-client'
 {EventEmitter} = require 'events'
 async          = require 'async'
 
@@ -15,11 +14,10 @@ class Basebot
   # dispatch them to matching listeners.
   #
   # adapter     - A String of the adapter name.
-  # httpd       - A Boolean whether to enable the HTTP daemon.
   # name        - A String of the robot name, defaults to Hubot.
   #
   # Returns nothing.
-  constructor: (httpd, name = 'Hubot', alias = false) ->
+  constructor: (name = 'Hubot', alias = false) ->
     @name       = name
     @events     = new EventEmitter
     @brain      = new Brain @
@@ -37,6 +35,7 @@ class Basebot
     @globalHttpOptions = {}
 
     @setupRouter(@nullRouter())
+    @setupHTTP(@nullHTTP())
 
     @errorHandlers = []
 
@@ -329,6 +328,21 @@ class Basebot
       delete: ()=> @logger.warning msg
     }
 
+  setupHTTP: (httpClient) ->
+    @http = httpClient
+
+  nullHTTP: ->
+    msg = "A script has tried making a HTTP request while the HTTP client has not been initialized."
+    return -> {
+      query: ()=> @logger.warning msg; ->
+      header: ()=> @logger.warning msg; ->
+      headers: ()=> @logger.warning msg; ->
+      get: ()=> @logger.warning msg; ->
+      post: ()=> @logger.warning msg; ->
+      put: ()=> @logger.warning msg; ->
+      delete: ()=> @logger.warning msg; ->
+    }
+
 
   # Load the adapter Hubot is going to use.
   #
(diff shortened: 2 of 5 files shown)


ajacksified:
An example of a programatically-created robot:

```
require('coffee-script/register');

var bot = require('hubot/src/basebot');
var shellAdapter = require('hubot/src/adapters/shell');

var startHTTPServer = false;
var name = 'robert';
var alias = 'r';

var Bot = new bot(startHTTPServer, name, alias);
Bot.loadAdapter(shellAdapter);

var example = require('./scripts/example');
Bot.loadScript(example);

var ping = require('./scripts/ping');
Bot.loadScript(ping);

Bot.run();
```

technicalpickles:
This is pretty interesting! I'll need to review a little more thoroughly, but a few initial impressions:

* I'm not in love with the Basebot name, but I don't have a better suggestion yet
* should Basebot include the httpd stuff? maybe it should stay in Robot and programatic users can add it in themselves
* should parseVersion be moved to Basebot so the version is available? It gets used by httpd at least, and scripts feasibly could check it too.

> There should be no user-facing changes to how hubot works.

It's like you know me :sparkling_heart: 

cc https://github.com/github/hubot/issues/858 for programaticly launched hubot

ajacksified:
* Yeah, 'Basebot' is a pretty awful name.
* Good point - I'll break out httpd the same way I broke out express, so you can put in superagent or fetch-isomorphic or whatever instead, if you wish.
* The only reason I didn't was because I was trying to decouple it from `fs`, which is unfriendly to browserify/webpack (kind of an extreme example, but it's so close to working anyways.) Hardcoding the version number to match what's in package seems like a bummer, too.

michaelansel:
I too am intrigued :smiley: Gotta find time to do an in depth review (probably tomorrow), but my initial thoughts as we embark on making an awesome-but-massive change:
- how much cleanup do we want to roll into this refactor? (e.g. I see you improving the constructor signature)
- should Robot be the official shim layer for Hubot v2 and Basebot be API v3?
- for all new interfaces, I would love to formalize and test the snot out of them. Similarly, I'd like to explicitly call out what is *not* part of the interface so we can make future changes without guessing.
- @ajacksified, when we put together middleware (#724, #801, #803), it took...a bit of time. Hopefully we can move a little faster this time around, but I just want to set expectations appropriately :-)

I **really** like the idea of using this as an opportunity to provide v3 awesomeness without breaking all existing v2 bots. Double interfaces obviously isn't something we want to support forever, but lines up well with our previous thoughts around releasing all the new functionality of v3 and then deprecating v2 (can't find the comment, but I think you'll recall the conversation, @technicalpickles).

ajacksified:
Understanding that those notes may have been meant for _the owners_, not for me, my thoughts:

* I was trying to not clean up _too_ much - touching as little as absolutely necessary to make it possible to load something programatically. I'll leave the hefty refactoring for v3... unless you want to refactor a whole lot more. I'm happy to do more, but I'm afraid of breaking backwards compatibility.
* I very much like the idea of robot-as-a-shim. Again, backwards compatibility woo
* Good call. I probably should break up the tests as well.
* Totally understand! I can use my fork for my nefarious purposes in the meantime.

v3 is going to be ES6+, right? :trollface:

michaelansel:
Did some code reading and I think if we are going to split, now is probably the right time to start tearing things out that we don't want in the core Basebot interface. I know this expands the scope of the change, but the moment we expose Basebot as a thing, we will need to support everything it does.

Did some code reading, and I think we should make Basebot a very carefully considered interface (based on everything that has been learned since Hubot v2 was first released). I know this expands the scope of the change, but if we just carry over bad decisions to the Basebot interface, we haven't really gained much more than yet another interface to test and support. Instead, I think we should start with a minimum functional interface for Basebot and expand out as we see fit; everything else needed for current backwards compatibility goes into Robot.

So, here is a starting suggestion of which functions should be in which class:

Robot
- All the commonly modified internal variables: `@commands`, `@listeners`, more? Backwards compatibility with [getters and setters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty).
- `robot.setupHttp`, `robot.http`
- `robot.version`
- `robot.setupRouter`, `robot.router`
- `robot.hear`, `robot.respond`, `robot.topic`, `robot.enter`, `robot.leave`
- `robot.send`, `robot.reply`, `robot.messageRoom`
- Anything that does a `require` (`loadAdapter`, `loadScript`, et al)
- Documentation/Help parser
- Brain (move to Basebot once we pick a v2 interface)

Basebot
- Message Processing Core
  - `robot.listen`, `robot.catchAll`
  - Ability to add/remove arbitrary listeners
  - `robot.receive`
  - `robot.listenerMiddleware`, `robot.receiveMiddleware`
- Plugin interface
  - `robot.loadAdapter` - just `adapter.use this`? Or should we make this even more generic? I'm not yet convinced `Adapters` are really special as far as Basebot is concerned (other than setting `robot.adapter`).
  - Ability to add/remove commands/help information (or should this just be part of the Help plugin?)
- Response/reply
  - `robot.responseMiddleware`
  - `Response` object
  - I dream of someday changing the `Response` interface, which would suggest that it should only be in `Robot`, but that would also require changing the `Adapter` interface, so maybe that is unrealistic right now...
- Utilities
  - `robot.error`
  - `robot.run`, `robot.shutdown`

Anything that we hack into `Robot` (e.g. Brain, Help, Express, etc.) should be somehow reusable for Basebot (even if it is just replicating the same hacks that are in Robot). The idea here is that you can relatively easily use Basebot and bolt on whatever extra functionality you want to make it look like `Robot` until we get all the desired core functionality ported over.

simonbuchan:
Basebot should obviously be called [Lobot](http://starwars.wikia.com/wiki/Lobot) 😺 (Cause it's **lo**w-level, geddit?)

PR #1135, opened 2016-02-23
Updated the 'Creating a Script Package' docs to link to the 'Testing …

MonicaG:
…Hubot Scripts' section


technicalpickles:
Good idea :smile: I'm not sure why I didn't do that in the first place.

Issue #1143, opened 2016-03-13
Content of brain.users


pchaigno:
Quick question: What is the intended content for [brain.users](https://github.com/github/hubot/blob/master/src/brain.coffee#L94-L98)?

0. A list of all users currently connected?
0. A list of all past and current users? 

This will help write adapters with the correct expected behavior.
Thank you.

technicalpickles:
When hubot encounters a message from a user, the user added to the brain, so not quite either. The effect would be a list of all past and current users though if hubot has been running.

Example:

* https://github.com/github/hubot/blob/master/src/adapters/campfire.coffee#L60 calls brain.userForId
* https://github.com/github/hubot/blob/master/src/brain.coffee#L100-L113 finds the existing user or adds one

portgasd666:
@technicalpickles but of the hubot uses some kind of persistence such as the redis-brain then the brain has all the users ever encountered across various runs of the hubot correct ?

technicalpickles:
To step back a second:

* without an scripts like redis-brain, `brain.users` would only contain users encountered so far of this run of hubot (ie the specific process)
* with a script like redis-brain
  * the script loads the brain at startup, so previously encountered users would be available
  * new users encountered are added to `brain.users`
  * the script flushes the entire `brain` to whatever is backing it

portgasd666:
That clears things up for me. Thanks a lot

pchaigno:
Thanks a lot for the explanation @technicalpickles!
Do you think I should document this somewhere (and open a PR)? If so, does it make more sense as a code comment or in `scripting.md`? This issue might be sufficient though...

technicalpickles:
There is already https://github.com/github/hubot/blob/master/docs/implementation.md for implementation details. Maybe a separate doc for persistence, or part of that one. A PR for documentation would be better than referencing a comment on a PR.

pchaigno:
Closing in favor of #1151.

PR #1151, opened 2016-03-20
Document persistence mechanisms

pchaigno:
As discussed in #1143, this pull request documents `hubot.brain`, and, in particular, the content of `hubot.brain.users`.

I'll be happy to update it based on comments; especially if the wording can be made clearer :smiley:

### DIFF
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -35,11 +35,11 @@ There are two primary entry points for middleware:
 
 ## Persistence
 
-Hubot has an in-memory key-value store exposed as `robot.brain` that can be used to store and retrieve data.
-Furthermore, `hubot-redis-brain` enables persistence across Hubot restarts.
-This package requires a Redis server.
+Hubot has a memory exposed as the `robot.brain` object that can be used to store and retrieve data.
+Furthermore, Hubot scripts exist to enable persistence across Hubot restarts.
+`hubot-redis-brain` is such a script and uses a backend Redis server.
 
 By default, the brain contains a list of all users seen by Hubot.
 Therefore, without persistence across restarts, the brain will contain the list of users encountered so far, during the current run of Hubot.
 On the other hand, with persistence across restarts, the brain will contain all users encountered by Hubot during all of its runs.
-This list of users can be accessed through `hubot.brain.users` and [other utility methods](https://github.com/github/hubot/blob/master/src/brain.coffee#L94-L136).
+This list of users can be accessed through `hubot.brain.users()` and other utility methods.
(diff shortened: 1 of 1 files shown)


docs/implementation.md
```diff
@@ -32,3 +32,14 @@ There are two primary entry points for middleware:
 
 1. `robot.listenerMiddleware` - registers a new piece of middleware in a global array
 2. `middleware.execute` - executes all registered middleware in order
+
+## Persistence
+
+Hubot has an in-memory key-value store exposed as `robot.brain` that can be used to store and retrieve data.
```
technicalpickles: "in-memory key-value" store may be misleading. When you say key-value, that typically suggets something like redis. In reality, it's just a javascript object, which technically has keys and values.

docs/implementation.md
```diff
@@ -32,3 +32,14 @@ There are two primary entry points for middleware:
...
+Furthermore, `hubot-redis-brain` enables persistence across Hubot restarts.
```
technicalpickles: I would say that there are hubot scripts which enables persistence, and list hubot-redis-brain as one example.

docs/implementation.md
```diff
@@ -32,3 +32,14 @@ There are two primary entry points for middleware:
...
+This package requires a Redis server.
+
+By default, the brain contains a list of all users seen by Hubot.
+Therefore, without persistence across restarts, the brain will contain the list of users encountered so far, during the current run of Hubot.
+On the other hand, with persistence across restarts, the brain will contain all users encountered by Hubot during all of its runs.
+This list of users can be accessed through `hubot.brain.users` and [other utility methods](https://github.com/github/hubot/blob/master/src/brain.coffee#L94-L136).
```
technicalpickles: `hubot.brain.users` is actually [a function](https://github.com/github/hubot/blob/master/src/brain.coffee#L94-L98), so I'd include parens on it ie `hubot.brain.users()`.

I also wouldn't link to line numbers in brain.coffee in documentation either, because that is a good way for it to become out of date.

pchaigno: Make sense, I removed it.

docs/implementation.md
pchaigno: Changed it :smiley: 
I actually copied this first line from [`scripting.md`](https://github.com/github/hubot/blob/master/docs/scripting.md#persistence), so maybe I should update it too? I'm wondering if we should explain in `scripting.md` that `get` and `set` are only needed for `hubot-redis-brain`. Maybe in another pull request?

Issue #1153, opened 2016-03-21
Loading .env file from within hubot


twellspring:
Hubot currently uses environment variables to configure the hubot instance.  bootstrap files also show the ability to load from .env.  Example:

```
# Load environment specific environment variables
if [ -f .env ]; then
  source .env
fi
```

This works as long as the .env file has `export` in front of all the variables.  
`export HUBOT_SLACK_TOKEN=xoxb-xxxxxxxxxxx-xxxxxxxxxxxxxxxxxxxxxxxx`

Common node.js practice is to use modules like dotenv or node-env-file to load variables from .env in the format 
`HUBOT_SLACK_TOKEN=xoxb-xxxxxxxxxxx-xxxxxxxxxxxxxxxxxxxxxxxx`

But that format does not work with the `source .env` method above.  

Adding `node-env-file` to bin/hubot would resolve this issue and be compatible with current functionality

```
env      = require 'node-env-file'
env('.env', {verbose: false, raise: false})
```

This will 
- only load items in .ssh that do not have export in front of them ( so will not conflict with existing .env files being sourced )
- not overwrite any environment variables that are already populated via the environment

twellspring:
# Tests

## Gets read 
.env
```
HUBOT_SLACK_TOKEN=xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxx
```
**Results**
```
# bin/hubot -a slack
[Mon Mar 21 2016 10:00:41 GMT-0700 (PDT)] INFO Connecting...
[Mon Mar 21 2016 10:00:42 GMT-0700 (PDT)] INFO Logged in as hubot of xxxxx, but not yet connected
[Mon Mar 21 2016 10:00:43 GMT-0700 (PDT)] INFO Slack client now connected
```

## Not read if export 
.env
```
export HUBOT_SLACK_TOKEN=xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxx
```
**Results**

```
# bin/hubot -a slack
[Mon Mar 21 2016 10:03:00 GMT-0700 (PDT)] ERROR No services token provided to Hubot
```

## Will not overwrite
bin/hubot 
```
export HUBOT_SLACK_TOKEN=xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxx
exec node_modules/.bin/hubot --name "hubot" --alias "!"  "$@"
```
.env
```
HUBOT_SLACK_TOKEN=this-is-a-bad-token
```
**Results**
[...]

PR #1154, opened 2016-03-21
add node-env-file to hubot to parse .env files

twellspring:
Adding node-env-file to bin/hubot so it can read standard node .env files 
Reference:  issue #1153 

If someone can suggest where this should be documented I will add documentation changes to pull request.

### DIFF
--- a/bin/hubot
+++ b/bin/hubot
@@ -7,6 +7,8 @@ Hubot    = require '..'
 Fs       = require 'fs'
 OptParse = require 'optparse'
 Path     = require 'path'
+env      = require 'node-env-file'
+env('.env', {verbose: false, raise: false})
 
 Switches = [
   [ "-a", "--adapter ADAPTER", "The Adapter to use" ],
--- a/bin/hubot
+++ b/bin/hubot
@@ -7,8 +7,7 @@ Hubot    = require '..'
 Fs       = require 'fs'
 OptParse = require 'optparse'
 Path     = require 'path'
-env      = require 'node-env-file'
-env('.env', {verbose: false, raise: false})
+env      = require('dotenv').config({silent: true}); 
 
 Switches = [
   [ "-a", "--adapter ADAPTER", "The Adapter to use" ],
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "connect-multiparty": "^1.2.5",
     "express": "^3.21.2",
     "log": "1.4.0",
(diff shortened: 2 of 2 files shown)


gasolin:
maybe it's better to also ignore .env file on `.gitignore`, so people wont check it in accidentally?

dnrce:
What's your reasoning for choosing `node-env-file` (v0.1.8, 9246 downloads this month) over `dotenv` (v2.0.0, 249627 downloads this month)?

twellspring:
Yes using dotenv is a better choice.  I have updated it to use dotenv.   Functionality is the same.

bhuga:
> maybe it's better to also ignore .env file on .gitignore, so people wont check it in accidentally?

I like this thinking.

> If someone can suggest where this should be documented I will add documentation changes to pull request.

I think adding a quick line or to to the deployment platforms that this applies to is the right call.

If we want to make `dotenv` a first-class citizen, we should probably also generate it when setting up a new hubot.

Ryangr0:
Does this have any estimated merge date? I use this already on every implementation of hubot and manually edit the hubot file now, but that's obviously a very dirty way of doing this.

jd-carroll:
What's the status of getting this merged?

technicalpickles:
I'm not sure I like including this as a first-class citizen for hubot, just because people are going to manage environments their own way. I don't know what side effect this would have on existing users that manage their own stuff.

I think in the long run, I'd prefer to see movement towards programatic launching of hubot ie https://github.com/github/hubot/issues/858 , and generating a new hubot could generate the code to initialize a new hubot, with a one-liner to load with dotenv.

davidyell:
So this requires an extra dependancy?

bin/hubot
```diff
@@ -7,6 +7,7 @@ Hubot    = require '..'
 Fs       = require 'fs'
 OptParse = require 'optparse'
 Path     = require 'path'
+env      = require('dotenv').config({silent: true});
```
gr2m: you don’t need to set the `env` variable. I would also suggest to move up `require('dotenv').config({silent: true});` to the very first code line, as suggested in https://www.npmjs.com/package/dotenv#usage

gr2m:
sorry I missed the whole discussion before my review. I agree that if we should ever agree on merging this, we should add `.env` to the generated `.gitignore` and document how to sent ENV variables. 

I still have to review #858

PR #1155, opened 2016-03-22
Aspirational middleware scripts

chittunoo:


docs/patterns.md
```diff
@@ -142,3 +142,131 @@ Example access policy:
...
+## Middleware Examples
+
+Middleware is often used for authentication and authorization.  There are three kinds of middleware:
+- **Receive**: executed for every message prior to testing for listener matches
+- **Listener**: executed for every listener that matches a message before executing the listener callback
+- **Response**: executed for every message hubot attempts to send, regardless of reason or source
+
+Middleware is asynchronous by default, needing to call either `middleware.next()` to continue processing or `middleware.done()` to stop processing. There are Sync helper functions for each type to simplify creating simple middleware units.
+
+**All of these are just examples. None of them are secure. Don't use them.**
```
bkeepers: What would it take to create examples that people could copy-paste and use? Or could this comment at least explain what about them is insecure so people don't have the same issues with the middleware they create based on these examples?

michaelansel: I'm somewhat dredging this out of the depths of my memory, but I think the big issues here was having a good understanding of your threat model and the list of all listeners. Blacklisting is particularly dangerous if you are auto-updating scripts and the upstream changes the listener ID (it is suddenly allowed with the same regex).

I ended up rewriting this section when we actually merged ("Restricting access to commands") -- https://hubot.github.com/docs/patterns/

michaelansel:
I thought we ended up merging all of these ideas in separate PRs. @bhuga can you confirm?

bhuga:
> I thought we ended up merging all of these ideas in separate PRs. @bhuga can you confirm?

Yeah, we discussed this and we decided that we wanted a lot of it to be in `scripting.md` instead of `patterns.md`. The actual auth example in this PR made it into `patterns.md` somehow or another. I think we can close this one out.

PR #1166, opened 2016-04-12
Add link to Facebook Messenger Platform adapter

MrSaints:
This is an adapter for Facebook's new Messenger Platform, and its Send / Receive API.


technicalpickles:
The only thing I'd suggest is to include the name `facebook` in the adapter name for clarity and to avoid confusion. I can merge as is, but let me know if you make any changes to the name in the form of a PR 😁

PR #1209, opened 2016-06-30
Make shell history file secure by default

denisvm:
Change the default file mode from `0o666` (u=rw,g=rw,o=rw) to `0o600` (u=rw,go=), which will deny access to everyone else other than the owner of the file.
This is used only when creating the file, so the user can change it afterwards if really want to make it public.


bkeepers:
Thanks @denisvm. We'll cut a new release soon that has this fix in it.

PR #1225, opened 2016-08-07
Fix broken link for Getting Started

devashishmamgain:


bkeepers:
#1309 will fix all links so they work on github.com and hubot.github.com.

PR #1239, opened 2016-09-21
Replace log with npm log

timkinnane:
Makes logs easier to read, more consistent with npm logs for shell debugging.
Date formats are localised and styles allow more obvious log level and separation of headers/prefix/message from a glance.

Could also be extended to log [error stacks](https://github.com/npm/npmlog/blob/master/example.js#L34-L37) from within hubot for more transparent debugging of external script functions.

screenshot 2016-09-21 17 33 29

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: node_js
-
-matrix:
-  include:
-    - node_js: "0.10"
-    - node_js: "0.12"
-    - node_js: "4"
-    - node_js: "5"
+node_js:
+  - "node" # latest stable Node.js release
+  - "7"
+  - "6"
+  - "5"
+  - "4"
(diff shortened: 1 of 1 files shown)


technicalpickles:
@timkinnane hm, I think you force pushed so there isn't a diff anymore 🤔

technicalpickles:
Ah, looks like you may have reset your master to follow this master, with [log-style](https://github.com/timkinnane/hubot/tree/log-style) as the new one. A topic branch will definitely be useful if you are making multiple PRs 😁

timkinnane:
Yeah I did that to make another PR and keep my master tracking upstream. I couldn't even remember how this one worked, so I'll have to pull in master and update it in a couple days.

PR #1255, opened 2016-10-28
adds missing -> to code snippet

ksatirli:
Hi!

One of the code examples is missing a `->` and thus will not pass linting checks (or actually work).

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -481,11 +481,11 @@ Using previous examples:
     # rest of the code here
 
 
-  robot.hear /midnight train/i, (msg) ->
+  robot.hear /midnight train/i, (res) ->
     robot.http("https://midnight-train")
       .get() (err, res, body) ->
         if err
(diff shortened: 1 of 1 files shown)


docs/scripting.md
```diff
@@ -481,11 +481,11 @@ Using previous examples:
     # rest of the code here
 
 
-  robot.hear /midnight train/i, (res)
+  robot.hear /midnight train/i, (msg) ->
```
technicalpickles: This should remain `res` to match other documentation and that it's a `Response` object

docs/scripting.md
```diff
@@ -481,11 +481,11 @@ Using previous examples:
...
     robot.http("https://midnight-train")
       .get() (err, res, body) ->
         if err
-          res.reply "Had problems taking the midnight train"
+          msg.reply "Had problems taking the midnight train"
```
technicalpickles: This should remain `res` too.

PR #1256, opened 2016-10-30
rm extension

khutchi:
make it look like a typical license file without extension


technicalpickles:
Any particular reason, or just for consistency?

PR #1286, opened 2017-01-20
Clearer README on what hubot really is

mose:
.. and removing the link to webchat will avoid misleading visitors. In the past 6 months, people just jump in, expect to be able to test a bot, and don't stay. But there is no bot on that irc channel, as it's more a social place for developers ...

I hope this change in the README could improve the quality of our irc channel :)


miklb:
As someone who idles in that channel, I echo this sentiment.

bkeepers:
Thanks!

PR #1290, opened 2017-01-27
Fix list numbering in Gotchas-section

maxkoryukov:
The numeration in the **Gotchas** section was broken, because of misformatted code-samples.


bkeepers:
Thanks!

PR #1303, opened 2017-03-21
Configurable limit on JSON payload size

sirtow:
Sometimes the default 1mb limit on json body is not enough. 
Adding the option to overide default by setting env variable `EXPRESS_REQUEST_SIZE`


bkeepers:
Actually, looks like this was already submitted in https://github.com/github/hubot/pull/1106, so closing this as a duplicate and we'll get that other one merged. Thanks for your contribution.

PR #1311, opened 2017-04-30
X

RahmadaniBayuAji:


strugee:
What are you trying to do here?

technicalpickles:
It looks like it was pulling in @ErinCall's changes over in https://github.com/github/hubot/tree/brain-mutation-log

This user has since been marked as spammy, and thus this PR too.

PR #1320, opened 2017-05-10
Curate list of adapters

bkeepers:
This cuts the list of adapters down to just the most popular services, and then links to a [GitHub topic search for `hubot-adapter`](https://github.com/search?q=topic%3Ahubot-adapter&type=Repositories) for the full list.

I'll open an issue on all the adapters to ask them to add the topic to their repository.


bkeepers:
> At one point, I was looking at making it NPM based

Good call. I added a link to the NPM search too since the topic list doesn't include everything for now.

Issue #1294, opened 2017-02-12
Scripting Patterns link in docs is giving 404


itomaldonado:
Some of the pages in the Hubot docs are linking to the url seen below and it does not exist (404 error).

`https://hubot.github.com/docs/scripting/patterns/`

mose:
You are right indeed, but I found only one such erroneous link. Thanks for reporting!

mose:
Actually the link was working. If you can find precisely where the link is broken, can you report it again? Maybe it has been actually fixed in the while. Thanks anyways :)

PR #1325, opened 2017-05-12
fix link to patterns

mose:
tiny fix noticed on issue #1294

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -305,7 +305,7 @@ module.exports = (robot) ->
   )
 ```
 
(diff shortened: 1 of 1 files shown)


bkeepers:
The link looks like it's working right now on [docs/scripting.md](https://github.com/github/hubot/blob/master/docs/scripting.md) and https://hubot.github.com/docs/scripting/. This change breaks it for me.

mose:
Indeed. hmm, then there is nothing to do, lets close that.

PR #1328, opened 2017-05-12
Add a new / updated generator to the roadmap

technicalpickles:
It seemed like a good idea at the time, but the generator at https://github.com/github/generator-hubot is a bit of a maintenance headache. Previously, it was just part of the hubot package and just dropped files on disks without a 'real' generator.

https://github.com/github/generator-hubot/issues has some of the current issue, but I have also seen in the past lots of issues with just installing `yeoman`, since it has a ton of dependencies.

I'm not picky about where on the roadmap it is, but it should go somewhere.

cc @bkeepers @mose


mose:
Yes I think we really need to add it on the roadmap.

PR #1329, opened 2017-05-13
Adding a helpful tip for Slack integration

jbosse:
This solution can be found on SO but it might be nice to have it here.


mose:
Nice, thanks :)

PR #1336, opened 2017-05-20
Update node versions that Travis runs against

bkeepers:
It looks like the Travis builds got deactivated at some point. I turned them back on, and while I'm at it, am updating the versions of node that it tests against.

Changes:

- ~~Drop `0.10` (`0.12` is still supported)~~
- Add `6`, `7`, and `node` for latest stable release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: node_js
-
-matrix:
-  include:
-    - node_js: "0.10"
-    - node_js: "0.12"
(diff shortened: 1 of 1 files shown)


mose:
well, debian stable (jessie) still ships with 0.10 :(

cf https://packages.debian.org/jessie/nodejs

bkeepers:
Ok, updated to keep `0.10` until we make a conscious decision to drop it.

Issue #1279, opened 2016-12-26
Promises in receiveMiddleware


javiersuweijie:
I want to intercept all incoming messages and send them to an API to do some pre-processing. 

```
robot.receiveMiddleware (context, next, done) ->
  message = context.response.message
  if message.constructor.name is "TextMessage"
    API.call(message)
      .then (data)->
        next(done)
      .catch (error)->
        done()
```

`next(done)` does not trigger the other listeners in the queue. Is there a better way to do this?

PR #1339, opened 2017-05-22
feat(middleware): Return a promise that resolves when middleware stack done

### LINKED ISSUE #1338: Let middleware return a promise
Hey humans. I keep hitting a logic wall where I need to wait for hubot to finish processing middleware, but the only way to do that currently is to add more middleware, which doesn't work with my requirements. That approach could also be buggy, as you'd still want to resolve the promise even if the first middleware prevents others from executing.

### Use case 1:

Unit testing expects `res.reply` strings to be transformed by middleware.
Currently I have to wait and observe the test helper's messages for a matching pattern, calling the test's done when the reply middleware stack finishes and finally outputs a message, because res.reply returns immediately but the transform happens asynchronously. Very pain.

If middleware.execute returned a promise, it would cascade up through `runWithMiddleware` and all the response methods that invoke it, like send, emote, reply etc. Much wow.

Then I could very simply do something like this:

```
it 'capitalises a string', (done) ->
  res.reply 'sent string'
  .then (context) ->
    expect(context.strings).to.equal ['Sent String']
    done()
```

### Use case 2

after a user sends something to hubot, listener middleware stores the match and performs some lookup for correlating data. I need to use that data in the listener callback. I can't move the middleware logic to the callback (because reasons), but I can't reply until I know the middleware has completed.

I've heard that async logic support will be improved in the next version, but this is a pretty small change that will really increase the versatility of my bots, so I'm hoping I can get some help with a PR and get it through without waiting for broader roadmap stuff.

I'll post PR to follow up, but the middleware logic looks like black magic to me and I'm unfamiliar with the async library, so it won't be accepted without feedback. Thanks in advance.

timkinnane:
Fixes #1338 
Related #1279

Needs help / feedback...
- The sinon fake timers don't seem to effect setTimeout, so it's actually taking the full half second to run the tests.
- How should I reject the promise if the middleware is cancelled or throws?
- I can change inline comments with new return values, but how should I update published docs?


timkinnane:
Just noticed Travis failed, but only on Node 0.1.0 due to lack of promise support.
How should I resolve that? I could use bluebird to polyfill, but I'm not sure about the etiquette of adding deps with PRs. Or is it time to ask if node <1 support is still required? Especially with new roadmap features coming up that might have similar support requirements.

mose:
Hubot 3 will certainly have new node version requirement, but there is still many 0.10 out there (cf https://github.com/github/hubot/pull/1336), I think it would be bad to introduce a breaking change for them at this point.

timkinnane:
Aha, fair play. What then? Can I add a dependency then - e.g. [promise](https://www.promisejs.org/) or [bluebird](bluebirdjs.com)?

bkeepers:
@timkinnane thanks for the pull request! I think we should wait until the next major release to drop support for v0.10.

In the mean time, I'm not opposed to adding a dependency so we can start using promises now. [es6-promise](https://www.npmjs.com/package/es6-promise) would be my preference so we can use the native `Promise` on every other version of node.

```coffeescript
Promise = global.Promise || require('es6-promise').Promise
```

technicalpickles:
> I can change inline comments with new return values, but how should I update published docs?

You can commit changes to `docs`, and those will automatically (or should) get updated on hubot.github.com

src/middleware.coffee
```diff
@@ -1,4 +1,5 @@
 async = require 'async'
+Promise = global.Promise || require('es6-promise').Promise
```
technicalpickles: I think https://github.com/stefanpenner/es6-promise#auto-polyfill means you could require the auto-polyfill, and then just use `Promise`. On the other hand, using this style is a lot more clear on where Promise is coming from.

technicalpickles:
> How should I reject the promise if the middleware is cancelled or throws?

I think you'd need to get promises involved a little earlier in the `execute`. The way `executeSingleMiddleware` currently works, it either succeeds, or it emits an error and calls done.

Maybe if you wrap more of the method in the promise call, you'd be able to resolve/reject during those checks?

timkinnane:
Sweet, I've made those changes and added docs.

I added one test that fails, because I need some advice to fix it.
How can I make the promise resolve early if a middleware calls done, before final done?

test/middleware_test.coffee
```diff
@@ -238,7 +296,7 @@ describe 'Middleware', ->
             {}
             middlewareFinished
             middlewareFailed
-          )
+          ).catch (reason) -> # supress warning re unhandled promise rejection
```
bkeepers: Will this start showing unexpected warnings when people upgrade?

timkinnane: Without that line 299, it was logging a warning in the tests, where middleware threw. Saying "unhandled promise rejection" - not a big deal, but simply catching the promise and doing nothing with it removes the warning. It would only effect people who's middleware is also throwing, in which case they'd see errors anyway.

timkinnane:
@bkeepers yeah, so any middleware in the stack can prematurely end it, but that's not really a fail, so the promise should still resolve with the context at that point. I just can't see how to hook in the resolve method at that point. The way I've done it only works through the final done function. 

The test I added at line [240](https://github.com/github/hubot/pull/1339/files#diff-df7bbfb83aaa129a0a4bd1e5aff03439R240) demos how it *should* work. I'm sure I'll be able to fix this myself, just wanted throw it up in case someone has a simple solution.

test/middleware_test.coffee
siorikitajima: totally

timkinnane:
Fixed the test I added, for resolving promise early if middleware calls done()

src/middleware.coffee
```diff
@@ -21,32 +22,44 @@ class Middleware
...
+        nextFunc = (newDoneFunc) -> cb(null, newDoneFunc or doneFunc)
+        # Catch errors in synchronous middleware
+        try
+          middlewareFunc.call(undefined, context, nextFunc, doneFunc)
+        catch err
+          # Maintaining the existing error interface (Response object)
+          @robot.emit('error', err, context.response)
+          # Forcibly fail the middleware and stop executing deeper
+          doneFunc()
+          reject err, context
```
gr2m: `Promise.reject()` only accepts one argument, you can try it out using

```js
Promise.reject(new Error('foo'), {}).catch(function (error, context) {
  console.log(context)
})
// logs `undefined`
```

What you could do is set context as property of `err` before rejecting with it

docs/scripting.md
```diff
@@ -702,6 +702,10 @@ Every middleware receives the same API signature of `context`, `next`, and
 `done`. Different kinds of middleware may receive different information in the
 `context` object. For more details, see the API for each type of middleware.
 
+Middleware execution returns a promise that resolves with the final `context`
+when the middleware stack completes. If the middleware stack throws an error,
+the promise will be rejected with the error and `context` at that point.
```
gr2m: see my other comment, a promise can only reject with one argument

gr2m:
hey @timkinnane I gave your PR finally a more thorough look today.

One thing that really confuses me is this part:

> middleware in the stack can prematurely end it,

I try to wrap my head around it but don’t understand why that is. Could someone explain it to me? Is it so that a there can be an array of registered middleware which all do their own checks and if one of them finds that the request is for them, then we don’t want the next middleware to even do its check?

timkinnane:
So normally it would run all middleware methods in the order they're registered, then a final function when complete, but each middleware in the queue can end the processing of the stack. The result of finishing early is that the "final" function never runs, but it isn't necessarily an error, just a different kind of conclusion. e.g. A listener middleware might prevent further middleware from processing if the user isn't authorised to access the listener, then reply with some "you are not valid" message instead of proceeding to the response callback. 

In such cases the promise should still resolve even though the stack was "cancelled" because `middleware.execute` still completed without throwing and you might want to write asynchronous logic that will interrogate the context of the middleware, wether it was authorised or not.

I'm currently away and will be for another week, but I'll put more work into this when I get back, to help it get through with a more consistent approach if required.

src/middleware.coffee
timkinnane: Good one. Thanks for that info, I'll do it that way instead.

gr2m:
thanks for the clarifications! Note that we will release a new major version of hubot soon which is now based on JavaScript and has a few minor breaking changes. But if for some reason you can’t upgrade to it we can release another `2.x` release with the new feature, given we agree that we want to add it

timkinnane:
Cheers, I've written *a lot* of code for 2.x and the promise feature is most useful for me when running mocha tests of those scripts. So I'd rather have it available without needing to update those scripts, but I'll look to evolve my packages to work better with v3 over time.

timkinnane:
Made a couple new branches to resubmit this separately for v2 and v3

PR #1340, opened 2017-05-23
Update npm-scripts (#978)

bkeepers:
This fixes some of the suggestions in #978.


package.json
```diff
@@ -43,6 +43,14 @@
     "hubot": "./bin/hubot"
   },
   "scripts": {
-    "test": "./script/test"
+    "clean": "rm -rf lib/",
+    "preinstall": "script/dotenv",
```
technicalpickles: Do we know if using `pre<thing>` works for these? If npm forks to run these, I don't think the changed environment would get back to the npm process that then calls the `<thing>`. My impression is these are more for like if you needed to run `./configure` which would create some files and whatnot.

bkeepers: I just tested and it looks like it doesn't. I'm guessing you're right that it forks. 😞 

It looks like `.env` was only loaded from `script/bootstrap` before. Do you now what this was used for?

#1154 was proposing adding `.env` support for all of Hubot, which seems like something we might want to consider.

I can work on a fix for this tomorrow. Let me know if you have any idea what bootstrap was using .env for.

technicalpickles: I'm not sure exactly.

I think at one point, you could just run hubot out of a clone of this repository. That introduced some problems where people would fork this (the framework) repository, and put their scripts in it, rather than using it as a framework. I think that was undone at one point, but maybe `.env` here is an artifact of that?

PR #1342, opened 2017-05-24
Fixed issue with adapterPath not being set correctly in Robot constru…

### LINKED ISSUE #1297: Cannot set custom adapter path through constructor of Robot class
Hi,

I think there is an error here: https://github.com/github/hubot/blob/498d65611388d2881e6ae24f74bd68a832a24211/src/robot.coffee#L43-L44

Since in the constructor argument  for `adapterPath` is expected to be `nil` or existing, `@adapterPath` will never be existing since it is not declared anywhere (the constructor uses `adapterPath` and not `@adapterPath`).

rqc:
…ctor.

Fixes #1297

### DIFF
--- a/src/robot.coffee
+++ b/src/robot.coffee
@@ -34,15 +34,14 @@ class Robot
   # Robots receive messages from a chat source (Campfire, irc, etc), and
   # dispatch them to matching listeners.
   #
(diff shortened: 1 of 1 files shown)


rqc:
@bkeepers 👍

PR #1344, opened 2017-05-26
Remove old dotenv stuff

bkeepers:
Per discussion in https://github.com/github/hubot/pull/1340#discussion_r118395796, these files seem to be unused relics.


technicalpickles:
If we do need hubot loading .env, I'd consider something like https://github.com/github/hubot/pull/1154

Issue #1138, opened 2016-03-02
Port hubot from coffeescript to plain Javascipt (ES2015)


gasolin:
.coffee does not look familiar for JS developers. 
After es6 is on the table and its easy to make es6 work on modern JS runtime through babel. Developers will have less motivation to learn .coffee and might cause maintenance issue in long term.

We almost complete the porting from hubot coffeescript to es6 (through babel)
including all .coffee and test cases. CI test passed and even passed eslint style check.
All hubot plugins are still workable.

Since the code looks different now, I setup a separate repo:
https://github.com/gasolin/webbybot

@technicalpickles would you consider move hubot to es6 based?

gasolin:
We almost complete the porting . hubot-generated project just need minor change to use this es6 version. (mainly because I take different naming from hubot to better test the port result and avoid the confusion)

https://github.com/gasolin/webbybot#how-to-replace-hubot-to-webbybot

Still wait for feedback if hubot interesting in moving to es6 based.

gasolin:
Hi, @technicalpickles , do you like Javascipt version, or I should fork from the Hubot?

gasolin:
Wrote an article to explain the changes from Coffeescript to ES6 http://blog.gasolin.idv.tw/2016/03/how-we-ported-hubot-from-coffeescript.html

technicalpickles:
I'd consider moving to ES6, but I'd be worried about it be effectively a rewrite. If there's a way to change individual files over one at a time, that'd be a lot lower risk. The biggest consideration I have is that it continues to work for existing hubot users.

gasolin:
It's totally possible to change individual files over one at a time (that's what we did when porting hubot), hubot providing 77% test coverage, the ported result does fully compatible with current version, all plugins are still working.

I've created yeoman-generator-equivalent template to triage some edge cases, it works well
https://github.com/gasolin/webby_template

If that sounds good, I could start sending some patch to introduce the ES6 (babel) build script and change individual file one by one.

enapupe:
@gasolin I'd like to thank you for your effort. You really have a point here. I came to github just to find an issue about this subject (my actual google query was `hubot fork es6`).
I have no interest in coffescript and that is frustrating.. hubot would definitely be 10x more exciting if it was ES6..
Thanks for putting so much time into something like that, I hope it merges with github interests, it would be a big win for everyone.

rqc:
[...]

PR #1347, opened 2017-06-02
CoffeeScript 🔜 JavaScript

gr2m:
The [JavaScript proposal for Hubot Evolution](https://github.com/hubotio/evolution/pull/4#issuecomment-305557470) is still ongoing, so everything can change. If you have any questions better to discuss there. 

But at this point I have the source code working in JavaScript and all existing coffee script tests are passing, so I thought I’d be a good time to start sharing progress.

### 🛑 this PR is not yet ready for any feedback on the code

Most of it is still generated from `decaffeinate`. I’ll test the current version with some popular adapters / scripts first to makes sure it all still works

closes – eventually – https://github.com/hubotio/evolution/pull/4

### Breaking Changes

- node@0.10 and node@0.12 are no longer supported
- `require('hubot/src/adapter')` does no longer work. It was never officially supported, but now it won’t work at all. Use `require('hubot').Adapter` instead
- `Middleware.ticker()` has been removed. Use `process.nextTick()` instead

### Features

- We now export es2015 `class` declarations of all Hubot classes at `require('hubot/es2015')`. 

### Progress

See [JavaScript Evolution Proposal](https://github.com/hubotio/evolution/pull/4) for details

- [x] Convert source files from CoffeeScript to JavaScript
     see [evolution/4/javascript-src](https://github.com/github/hubot/compare/evolution/4/javascript-src) branch [Build Status](https://travis-ci.org/github/hubot)
- [x] Convert test files from CoffeeScript to JavaScript
      see [evolution/4/javascript-tests](https://github.com/github/hubot/compare/evolution/4/javascript-tests) branch, [Build Status](https://travis-ci.org/github/hubot)
- [x] [#](#package-json) update package.json
- [x] [#](#linting) add script for JavaScript linting
- [x] [#](#test) Test with popular projects depending on Hubot
- [x] Update `CHANGELOG.md`
- [x] review by other maintainers 👀 
- [x] squash the WIP commits
- [x] merge 🎉 

All steps will be part of the same PR but split into separate commits so people
can follow more easily.

#### Convert source files from CoffeeScript to JavaScript

Convert all source files from CoffeeScript to JavaScript with a tool like [decaffeinate](https://github.com/decaffeinate/decaffeinate), see also [github/hubot#1138](https://github.com/github/hubot/issues/1138).

Make sure the existing tests (still written in CoffeeScript) run against the new JavaScript.

Now go through each file and improve the code readability by hand as needed. From my experiences with converting Hoodie from CoffeeScript to JavaScript quite a lot of manual work will be required.

#### Convert test files from CoffeeScript to JavaScript

Same as the previous steps only for the tests files this time.

When finished, create a separate branch with the tests written in JavaScript but the source code still in CoffeeScript to assure integrity.

<a name="package-json"></a>
#### Update package.json

Remove all CoffeeScript related dependencies and tooling. Update [engines](https://docs.npmjs.com/files/package.json#engines)

Make sure that Hubot scripts can still be written in CoffeeScript. We want to break as few existing scripts as possible and many of them are written in CoffeeScript. We treat it as legacy support and will drop support for scripts written in CoffeeScript in future.

<a name="linting"></a>
#### Add script for JavaScript linting

The only tool I would like to introduce as part of this proposal is a linting tool which will be run as part of `npm test`.

I suggest [standard](https://standardjs.com/). We’ve been using it in all our projects at Hoodie and Neighbourhoodie since 2+ years and never looked back. It's a zero-configuration JavaScript linter.

<a name="test"></a>
#### Test with popular projects depending on Hubot

- [x] [slack](https://github.com/slackapi/hubot-slack)
- [x] [hipchat](https://github.com/hipchat/hubot-hipchat)
- [x] [irc](https://github.com/nandub/hubot-irc)

Instructions to test an existing adapter with this PR:

1. Check out the `evolution/4/javascript` branch of this pull request

   ```
   git clone git@github.com:github/hubot.git
   cd hubot
   git checkout evolution/4/javascript
   npm install
   npm link
   ```

2. Now link it into your test app using an example adapter

   Using the example of slack here:

   ```
   npm install -g yo generator-hubot
   mkdir my-awesome-hubot && cd my-awesome-hubot
   yo hubot --adapter=slack
   npm link hubot
   ```

3. Now start hubot

   ```
   NODE_PATH=`pwd`/node_modules HUBOT_SLACK_TOKEN=<your token here> ./bin/hubot --adapter slack
   ```

   replace `<your token here>` with your own bots token from slack


gr2m:
I’ve just tested it with the IRC adapter which is written in CoffeeScript and it seems there is still some problem with inheritance. [this.adapter.run()](https://github.com/github/hubot/pull/1347/files#diff-8591343033d9e39d1213cc5bc0ec5a2cR637) is the empty dummy `run () {}` from the abstract `/src/adapter.js` instead of the one from the irc adapter

gr2m:
okay, fixed via https://github.com/github/hubot/pull/1347/commits/9aa5fe71d57ca25fc1474d0391179b0952049237 

Tested it with the IRC adapter and it works. Doesn’t mean that it wouldn’t break with something else. This isn’t super elegant, but still our best option right now as we transition to move forward with the new roadmap and revive the project’s community I think :)

gr2m:
run into  a problem with `hubot-hipchat` thanks to @tombell, investigating now

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+## [v2.19.0](https://github.com/github/hubot/tree/v2.19.0) (2017-06-09)
```
bkeepers: > v2.19.0

Should this be v3.0?

bin/hubot
```diff
@@ -1,168 +1,9 @@
 #!/usr/bin/env coffee
-# vim:ft=coffee ts=2 sw=2 et :
-# -*- mode:coffee -*-
 
-Hubot    = require '..'
+# While all other files have been converted to JavaScript via https://github.com/github/hubot/pull/1347,
+# we left the `bin/hubot` file to remain in CoffeeScript in order prevent
+# breaking existing 3rd party adapters of which some are still written in
+# CoffeeScript themselves. We will depracate and eventually remove this file
+# in a future version of hubot
```
bkeepers: I'm not sure I understand this.  As long as it still includes a `require('coffee-script')`, shouldn't it be able to load JavaScript?

Also, should we print a deprecation warning that coffee script will not be supported in the major release that follows this one?

alangpierce:
Really cool to see this. 😄 If you have any feedback about how decaffeinate could be improved, feel free to let me know or file an issue.

gr2m:
@alangpierce decaffeinate was a great help! I compiled it with babel.js afterwards to target Node@4. It all wasn’t perfect, but I think the tooling is as great as it can be, thanks so much!

gr2m:
I’ve tested again with all 3: Slack, IRC and HipChat with my latest changes and they all continue to work

.travis.yml
```diff
@@ -1,12 +1,8 @@
 language: node_js
 node_js:
-  - "node" # latest stable Node.js release
```
gr2m: yeah it would lead to unexpected failures, node major versions are still breaking versions. We are experimenting with Greenkeeper sending PRs to update the these versions, so as soon as version 9 comes out it will send a PR updating the `.travis.yml` file. I think that’d be the perfect way

bin/hubot
gr2m: I don’t think `require('coffee-script')` will be enough, I'll have to investigate it. If you don’t mind I’d make a follow up PR to change it later if it works.

Regarding the deprecation warning I’m not yet sure when we drop CoffeeScript support officially, but we can add a deprecation warning in a patch release, too, so we don’t keep that PR waiting

cjdelisle:
I've ported the matrix adapter to js since this seems to be the evolution and I prefer js anyway, but I discovered an issue. I am getting the same `run() {}` issue as https://github.com/hubotio/hubot/pull/1347#issuecomment-307076144 and I'm not entirely sure how to deal with this. Apparently it now works well for coffee but not so much for js.

gr2m:
@cjdelisle Thanks for checking the PR with the matrix adapter! Do you have your code somewhere I can have a look? I thought I resolved the issue and tested it with both, adapters written in coffeescript as well as javascript.

cjdelisle:
Well I think it *worked* but I screwed it up porting to js. You can test it here https://github.com/xwiki-labs/xbot/blob/master/bot.js#L35 just do a `npm install`

gr2m:
I was able to reproduce the issue, I’ll look into it tomorrow

cjdelisle:
I'm gonna push a workaround for this and you can continue debugging with https://github.com/xwiki-labs/xbot/tree/c38054d4d6d3311a6232e0f353349d41342c043a
Thanks !

gr2m:
@cjdelisle it looks like you released a new version of `hubot-matrix` without pushing its code to GitHub is that possible?

The problem here is that you try to extend an es2015 `class` from a "CoffeeScript" class, this would not have worked before either. But what you can do now is to `require('hubot/es2015')` which gives you our internal es2015 class declarations (see https://github.com/hubotio/hubot/blob/master/es2015.js), then it will work.

Could you check that please?

cjdelisle:
everything is definitely pushed to github because I deleted `node_modules` and then ran `npm install` and then tested once finally before providing you the link.
I can confirm that the version of hubot-matrix which I'm using is ok with hubot-2.19 because I tested against this when converting hubot-matrix to js, perhaps the coffeescript loader sees es6 and adds some shim?
Requiring `hubot/es2015` is slightly annoying because I have to do it in a try/catch and fallback to the old behavior, otherwise break users with the old (current) version of hubot, but I can do this anyway. Thanks.

gr2m:
sorry I looked at a fork: https://github.com/xwiki-labs/hubot-matrix

Did you test hubot-matrix with the currently released version of `hubot` (hubot@2.19.0)? I don’t think that it would work. The problem is that es2015 classes cannot inherit from CoffeeScript classes and the other way around. The conversion you did here would not work: https://github.com/davidar/hubot-matrix/commit/d8224d3eb4f7ec9d139b745c1a7a3434114007e4#diff-44f2bbaefc4af762ff42533cf1f158d4R18

> Requiring hubot/es2015 is slightly annoying because I have to do it in a try/catch and fallback to the old behavior, otherwise break users with the old (current) version of hubot

yes that is true :/ But if you want to be compatible with the current `hubot` than you can’t use es2015 `class` inheritance anyway

What you can do is a to release a new version of hubot-matrix which would already require the new hubot version. We are coordinating the release of hubot@3 with the default scripts which get installed by the generator ourselves right now

cjdelisle:
This is what I tested and it seems to be working ok: https://github.com/xwiki-labs/hubot-test/blob/master/package.json#L15
Even if this is fine, I can propose @davidar another change to make it attempt to use 'hubot/es2015' before falling back.

technicalpickles:
If there's any followup work from this PR, I think it makes sense to track it in a new issue so we can track and 'close' it when it's done.

PR #1348, opened 2017-06-07
move MockAdapter into hubot repository

gr2m:
This is in preparation of #1347. Having the declaration of `MockAdapter` inside the Hubot repository makes sense anyway, it is fairly simple and we depend on it in our tests, so we should have full control. The main problem is that [`hubot-mock-adapter` does `require('hubot/src/adapter')`](https://github.com/blalor/hubot-mock-adapter/blob/5cb366c9150693624050bc2b04e39c83470de5e8/index.js#L23) which I don’t think is officially supported, it should be `require('hubot').Adapter` instead. 

If we want to keep a `hubot-mock-adapter` we should bring it up to date (e.g. it refers `src/scripts` in the README.md which doesn’t exist) and move it to the hubotio GitHub organization.

In #1347 we want to convert the tests independent of the source files, and the changes in this PR became necessary for the tests to still pass after converting all `/src/*` files to JavaScript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "scoped-http-client": "0.11.0"
   },
   "devDependencies": {
-    "hubot-mock-adapter": "^1.0.0",
     "coffee-errors": "0.8.6",
     "mocha": "^2.1.0",
     "mockery": "^1.4.0",
--- a/test/adapter_test.coffee
+++ b/test/adapter_test.coffee
@@ -4,7 +4,7 @@ chai.use require 'sinon-chai'
 
 expect = chai.expect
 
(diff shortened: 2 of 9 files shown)


mose:
It looks like it's working fine. I think that package still makes sense to be maintained for plugin authors, and having it revived in `hubotio/` sounds good. But I see no problem integrating it in main hubot code in any case.

gr2m:
Is the module used somewhere else? If we decide to move it to our `hubotio` org I’d be happy to use that again in the tests with a follow up PR, but would like to merge this first so we can move forward with #1347

PR #1352, opened 2017-06-13
parse help from script

gr2m:
I run into these problems while converting `hubot-help` from CoffeeScript to JavaScript. This is currently not covered by `hubot` tests. Having an end-to-end test that runs on CI would have caught that. I’ll look into it, but would like to merge this now if possible, as the change is minor

### DIFF
--- a/src/robot.js
+++ b/src/robot.js
@@ -516,8 +516,10 @@ class Robot {
   // Returns nothing.
   parseHelp (path) {
     const scriptDocumentation = {}
-    const body = fs.readFileSync(path, 'utf-8')
-    const lines = body.split('\n')
+    const body = fs.readFileSync(require.resolve(path), 'utf-8')
+
+    const useStrictHeaderRegex = /^["']use strict['"];?\s+/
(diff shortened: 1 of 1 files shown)


src/robot.js
```diff
@@ -516,8 +516,10 @@ class Robot {
   // Returns nothing.
   parseHelp (path) {
     const scriptDocumentation = {}
-    const body = fs.readFileSync(path, 'utf-8')
-    const lines = body.split('\n')
+    const body = fs.readFileSync(require.resolve(path), 'utf-8')
```
gr2m: before, people would pass paths with the file extension. As we want to be compatible with both JavaScript and CoffeeScript, we need to accept paths that can be required without the file extension. But for parsing the docs we need it again, which is exactly what `require.resolve` does.

src/robot.js
```diff
@@ -516,8 +516,10 @@ class Robot {
...
+
+    const useStrictHeaderRegex = /^["']use strict['"];?\s+/
```
gr2m: the `'use strict'` statement on top of the files is needed for Node@4 compatibility

PR #1353, opened 2017-06-17
Document using a custom listener to check for message text

technicalpickles:
Without this, any message is responded to, including someone entering or
leaving. Most common case is responding to text, so make sure to check
for that in the docs, and explain why.

Fixes https://github.com/hubotio/hubot/issues/1055


gr2m:
any idea why the `GitHub CLA` status check hangs?

technicalpickles:
I was able to fix the CLA check for this one, merging.

PR #1357, opened 2017-06-25
don’t load unsupported files

### LINKED ISSUE #1355: Javascript regression: Hubot tries to load all files in script directories, even if they aren't Javascript
@timothywellsjr posted on the [Javascript evolution](https://github.com/hubotio/evolution/pull/4) with a report that hubot was trying to require a YAML in `scripts`. @mistydemeo asked if the updated hubot is trying to load all files, not just javascript.

It seems like, yes, it is. Here's a dive through the relevant codes:

- bin/hubot tries to load `scripts`: https://github.com/hubotio/hubot/blob/master/bin/hubot.js#L115 
- `robot.load` finds all files in the `scripts` directory, then calls `robot.loadFile`: https://github.com/hubotio/hubot/blob/master/src/robot.js#L376
- `robot.loadFile` tries to require

Compare this to the previous behavior of `loadFile` which checks if it's a supported require type: https://github.com/hubotio/hubot/pull/1347/files#diff-740956d8fdad73632c4937d631b7c79bL353

cc @gr2m

gr2m:
closes #1355

### DIFF
--- a/src/robot.js
+++ b/src/robot.js
@@ -349,6 +349,11 @@ class Robot {
     const ext = path.extname(filename)
     const full = path.join(filepath, path.basename(filename, ext))
 
+    // see https://github.com/hubotio/hubot/issues/1355
+    if (!require.extensions[ext]) { // eslint-disable-line
+      return
+    }
+
     try {
       const script = require(full)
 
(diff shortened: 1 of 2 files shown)


gr2m:
@technicalpickles the GitHub CLA status is stuck again, can I ignore and just merge things in when there is a approving review?

bkeepers:
I disabled the CLA check for now.

FYI @mlinksva

PR #1358, opened 2017-06-25
update node/npm versions to LTS in "engines"

gr2m:
addresses https://github.com/hubotio/evolution/pull/4#discussion_r123839362

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,8 @@
     "standard": "^10.0.2"
   },
   "engines": {
-    "node": ">= 0.8.x",
-    "npm": ">= 1.1.x"
+    "node": ">4",
+    "npm": ">2"
   },
   "main": "./index",
   "bin": {


package.json
```diff
@@ -36,8 +36,8 @@
     "standard": "^10.0.2"
   },
   "engines": {
-    "node": ">= 0.8.x",
-    "npm": ">= 1.1.x"
+    "node": ">= 4.8.3",
```
bkeepers: Why 4.8.3, and not `>=4`? Just curious if there's something significant about 4.8.3, or if that's the latest 4.8 release.

gr2m: yes it’s the last maintenance release. But I think you are right, we should probably be fine with `>=4` and npm `>=2`, will show less errors. And if people run into problems with `4.x` versions less than 4.8 then we can bump that, I’ll update the PR

PR #1359, opened 2017-06-25
add tests for require(hubot/es2015)

gr2m:
part of https://github.com/hubotio/evolution/pull/4

### DIFF
--- a/es2015.js
+++ b/es2015.js
@@ -24,6 +24,6 @@ module.exports = {
   CatchAllMessage: Message.CatchAllMessage,
 
   loadBot (adapterPath, adapterName, enableHttpd, botName, botAlias) {
-    return new Robot(adapterPath, adapterName, enableHttpd, botName, botAlias)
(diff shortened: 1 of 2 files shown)


es2015.js
```diff
@@ -24,6 +24,6 @@ module.exports = {
   CatchAllMessage: Message.CatchAllMessage,
 
   loadBot (adapterPath, adapterName, enableHttpd, botName, botAlias) {
-    return new Robot(adapterPath, adapterName, enableHttpd, botName, botAlias)
+    return new module.exports.Robot(adapterPath, adapterName, enableHttpd, botName, botAlias)
```
technicalpickles: Was this not working previously?

gr2m: it worked, but this way we can stub `Robot` in tests 
https://github.com/hubotio/hubot/pull/1359/files#diff-0740e6d4a0c565b774aca19f820a7550R193

PR #1360, opened 2017-06-25
end-to-end test for "yo hubot"

gr2m:
part of https://github.com/hubotio/evolution/pull/4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,9 @@ node_js:
 notifications:
   email: false
 sudo: false
+addons:
+  apt:
+    packages:
+    - expect
+before_script: bin/e2e-test.sh
 after_success: npm run coverage
--- a/bin/e2e-test.sh
+++ b/bin/e2e-test.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+HUBOT_FOLDER=$(pwd)
+TEMP_ROOT=$(mktemp -d)
+
+echo "$ pushd $TEMP_ROOT"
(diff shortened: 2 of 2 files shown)


technicalpickles:
Huh, something about the node 8 environment really isn't liking the scoped-http-client

gr2m:
I’m looking into it

gr2m:
npm install /path/to/hubot will create a symlink in npm 5+. As the `require(scriptName)` calls for app-specific scripts happen inside hubot, we have to set NODE_PATH so they can be found when `hubot` is symlinked into the hubot app created with the generator. Easy workaround, I’ll push an update shortly

gr2m:
for the record, also had to change the order of `yo hubot` and `npm install $HUBOT_FOLDER` because `yo hubot` would delete the `node_modules` folder in `$HUBOT_FOLDER` 😮

gr2m:
hmpf :( something’s still not working :/ investigating

gr2m:
I can no longer reproduce the problem when running the `bin/e2e-test.sh` script locally. The difference is that on Travis `NODE_PATH` is undefined while it’s set correctly on my local machine, see https://travis-ci.org/hubotio/hubot/builds/247237160#L239

I guess there is some difference in how `expect` behaves differently between my mac and Debian on travis. I’m not sure how to work around this problem, I’d appreciate help from someone who is running linux locally and might have some experience with `expect` /cc @mose

gr2m:
💚 💚 💚  I think this ready to merge now

PR #1361, opened 2017-06-26
semantic-release setup

gr2m:
see

- https://github.com/hubotio/evolution/pull/6
- https://github.com/hubotio/evolution/blob/4/javascript/_drafts/javascript.md#publish-hubot300-with---tagnext


gr2m:
the test with Node 4 fails because npm errors if peerDependencies are not fulfilled while npm 3+ only shows a warning. I suggest we add `0.0.0-development` as a valid peerDependency, so the full value would be 

```json
  "peerDependencies": {
    "hubot": ">=2 <10 || 0.0.0-development"
  },
```

I tested it locally and it worked. I’ll update the evolution process and the PRs that I already sent to the scripts

gr2m:
@technicalpickles we will add a link to the last version of the `CHANGELOG.md` file on the 3.0.0 release notes

gr2m:
@technicalpickles alternatively we could create releases for past versions and put the changes in each release note, would you like me to do that?

PR #1366, opened 2017-07-06
Update Robot constructor to take an object of properties, rather than positional arguments

technicalpickles:
I started this as part of https://github.com/hubotio/evolution/pull/3 . The idea is to make the `Robot` a little easier to construct, so you don't have extra parameters that are unused/null (`adapterPath`), or boolean (`httpd`):

```javascript
// before
robot = new Hubot.Robot(null, 'shell', true, '/')

// after
robot = new Hubot.Robot({adapter: 'shell',  alias: '/'})
```

I updated `Robot` to take just the one `params`. It still works in the old way, but uses [depd](https://github.com/dougwilson/nodejs-depd) to print a deprecation warning. I'm open to just logging a message, but I was looking for something with decent output, and included line information. I was also thinking it'd make sense to have an 'upgrading' guide, that talks about specific things that need to be fixed, like this, and the deprecation could link back to it somehow.

I also deprecated `Hubot.loadBot` in favor of the constructor, since it doesn't seem to be doing a whole lot except constructing a bot. I could remove that if it it seems like that would be problematic.

TODO

- [x] write a test for using the old constructor arguments
- [ ] fix coveralls failure (probably addressed by the above?)
- [ ] cleanup commit history so semantic-release will like it

### DIFF
--- a/src/robot.js
+++ b/src/robot.js
@@ -7,6 +7,7 @@ const path = require('path')
 const async = require('async')
 const Log = require('log')
 const HttpClient = require('scoped-http-client')
+const deprecate = require('depd')('hubot')
 
 const Brain = require('./brain')
 const Response = require('./response')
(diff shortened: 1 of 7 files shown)


src/robot.js
```diff
@@ -21,11 +22,31 @@ class Robot {
...
+  constructor (params) {
+    var adapterPath, adapter, httpd, name, alias
+    if (arguments.length > 1) {
+      adapterPath = arguments[0]
+      adapter = arguments[1]
+      httpd = arguments[2]
+      name = arguments[3]
+      alias = arguments[4]
+
+      deprecate(`Robot constructed with multiple arguments, construct with one object argument instead`)
```
gr2m: just a minor style comment, I would use single quotes for strings unless you use place holders in it.

test/robot_test.js
```diff
@@ -28,8 +28,7 @@ describe('Robot', function () {
       warnOnUnregistered: false
     })
     mockery.registerMock('hubot-mock-adapter', require('./fixtures/mock-adapter'))
-    this.robot = new Robot(null, 'mock-adapter', true, 'TestHubot')
-    this.robot.alias = 'Hubot'
+    this.robot = new Robot({adapter: 'mock-adapter', httpd: false, name: 'TestHubot', alias: 'Hubot'})
```
gr2m: Even though it’s deprecated, I would suggest we keep a test for the old usage to make sure we don’t accidentally break it in future feature/fix releases. It will also increase the test coverage and make coveralls status pass :)

technicalpickles: That was on my TODO 😁

src/robot.js
```diff
@@ -21,11 +22,28 @@ class Robot {
...
-  // adapter     - A String of the adapter name.
-  // httpd       - A Boolean whether to enable the HTTP daemon.
-  // name        - A String of the robot name, defaults to Hubot.
-  constructor (adapterPath, adapter, httpd, name, alias) {
+  // @param {Object} params - parameters for building Robot
+  // @param {String} params.adapterPath -  A String of the path to built-in adapters (defaults to src/adapters)
+  // @param {String} params.adapter     - A String of the adapter name.
+  // @param {String} params.httpd       - A Boolean whether to enable the HTTP daemon.
+  // @param {String} params.name        - A String of the robot name, defaults to Hubot.
+  constructor (params) {
```
technicalpickles: I'm not super happy with `params` as the argument name. In Ruby, I'd call this options, but I'm not sure if there's a better node convention

src/robot.js
```diff
@@ -21,11 +22,28 @@ class Robot {
...
+    var adapter, httpd, name, alias
+    if (arguments.length > 1) {
+      // adapterPath = arguments[0]
```
technicalpickles: this isn't actually being used, but I figure it's useful to leave around to show what the first argument was.

technicalpickles: I figured I may as well fix https://github.com/hubotio/hubot/issues/1297 while here, I suppose 😅

src/robot.js
technicalpickles: some ~googling~ research suggestions `options` is a javascript thing to, ie http://www.codereadability.com/what-are-javascript-options-objects/

technicalpickles:
First attempt at using the commit message convention for semantic-release. Am I doing this right? 😅

gr2m:
That works :)

If you want to level up your commits game, you can create a separate commit for the package change, e.g. `chore(package): depd` and the test files: `test: Robot constructor to take an object of properties` and the new implementation with `feat: Robot constructor to take an object of properties`. 

This won’t make a difference for semantic-release though, it only cares about the `feat: ...` commit :) The benefit of having separate test & feat commits is that you can pus the test commit first so reviewers can directly see that the tests failed before they got fixed, as both will have a status from the CI

PR #1370, opened 2017-07-14
Middleware execute returns a promise (v2)

timkinnane:
Resubmitting PR #1339 with middleware.execute returning promise allowing async response methods - for v2 hubots, another PR coming for v3


timkinnane:
This has conflicts because changes were made to the 2.19.0 tag, before refactoring coffee to js for v3 release. I don't know if it's possible to make a PR to a tag, rather than a branch, but since the master branch is now dedicated to v3 development, can the hubot team create a branch at the final tag for v2 - for any further submissions to the v2 release in perpetuity? I would then base this PR to that branch.

PR #1371, opened 2017-07-14
Middleware execute returns a promise (v3)

timkinnane:
Updated my PR #1370 with v3 changes, including refactoring async middleware tests in js (from coffeescript) and I added some tests for response class, showing variations using / not-using returned promise.

There were no tests for the response class previous to this, so I'm pretty positive I haven't done it the best way, but I've at least provided a baseline for the team to review. Please give me feedback for how I can improve [those tests](https://github.com/timkinnane/hubot-async/blob/76b0d61313b0f17d9495de8c68cbfbb47f193433/test/response_test.js).


Dark403969 (approved):

joeyguerra:
We should probably use async/await instead.

Issue #1386, opened 2017-08-15
Hubot 3.0.1 robot.messageRoom with HipChat Adapter throws


mdouglass:
The following code when run using the HipChat adapter throws an error.
```javascript
robot.router.get('/hubot/test', (_req, res) => {
  res.status(200).json({})
  robot.messageRoom('119062_bot_testing@conf.hipchat.com', 'This is a test message')
})
```

Error:
```
TypeError: Cannot read property 'message' of undefined
  at Robot.HipChat.send (/opt/kixeye/20170815-235000000/node_modules/hubot-hipchat/src/hipchat.coffee:35:7, <js>:50:37)
  at Robot.messageRoom (/opt/kixeye/20170815-235000000/node_modules/hubot/src/robot.js:608:23)
  at robot.router.get (/opt/kixeye/20170815-235000000/scripts/router.js:13:11)
  at callbacks (/opt/kixeye/20170815-235000000/node_modules/express/lib/router/index.js:164:37)
  at param (/opt/kixeye/20170815-235000000/node_modules/express/lib/router/index.js:138:11)
  at pass (/opt/kixeye/20170815-235000000/node_modules/express/lib/router/index.js:145:5)
  at Router._dispatch (/opt/kixeye/20170815-235000000/node_modules/express/lib/router/index.js:173:5)
  at Object.router (/opt/kixeye/20170815-235000000/node_modules/express/lib/router/index.js:33:10)
  at next (/opt/kixeye/20170815-235000000/node_modules/connect/lib/proto.js:174:15)
  at Object.multipart [as handle] (/opt/kixeye/20170815-235000000/node_modules/connect-multiparty/index.js:47:63)
  at next (/opt/kixeye/20170815-235000000/node_modules/connect/lib/proto.js:174:15)
  at Object.urlencodedParser [as handle] (/opt/kixeye/20170815-235000000/node_modules/body-parser/lib/types/urlencoded.js:88:40)
  at next (/opt/kixeye/20170815-235000000/node_modules/connect/lib/proto.js:174:15)
  at Object.jsonParser [as handle] (/opt/kixeye/20170815-235000000/node_modules/body-parser/lib/types/json.js:100:40)
  at next (/opt/kixeye/20170815-235000000/node_modules/connect/lib/proto.js:174:15)
  at Object.query [as handle] (/opt/kixeye/20170815-235000000/node_modules/connect/lib/middleware/query.js:43:5)
[...]

PR #1390, opened 2017-09-09
Fix robot helpers delegation to adapter and events

slawekzachcial:
Robot 'send', 'reply', 'messageRoom' methods delegate invocations to
adapter but using as 'this' the robot while they should be using the adapter.
Robot 'on', 'emit' methods delegate invocations to events but using as
'this' the robot but they should be using the events.

The original CoffeeScript code from release 2.19.0 uses simple delegation, i.e. [send](https://github.com/hubotio/hubot/blob/c0f6bdf947c5271c2cac5d48730320b129c2e44f/src/robot.coffee#L542), [reply](https://github.com/hubotio/hubot/blob/c0f6bdf947c5271c2cac5d48730320b129c2e44f/src/robot.coffee#L552), [messageRoom](https://github.com/hubotio/hubot/blob/c0f6bdf947c5271c2cac5d48730320b129c2e44f/src/robot.coffee#L562), [on](https://github.com/hubotio/hubot/blob/c0f6bdf947c5271c2cac5d48730320b129c2e44f/src/robot.coffee#L573), [emit](https://github.com/hubotio/hubot/blob/c0f6bdf947c5271c2cac5d48730320b129c2e44f/src/robot.coffee#L583). This PR should address #1386.

### DIFF
--- a/src/robot.js
+++ b/src/robot.js
@@ -579,7 +579,7 @@ class Robot {
   send (envelope/* , ...strings */) {
     const strings = [].slice.call(arguments, 1)
 
-    this.adapter.send.apply(this, [envelope].concat(strings))
+    this.adapter.send.apply(this.adapter, [envelope].concat(strings))
   }
 
   // Public: A helper reply function which delegates to the adapter's reply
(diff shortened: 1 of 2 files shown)


seliver:
is there any ETA for these fixes to be released as part of a new version? 3.0.1 is unusable because of this.

diversario:
Yeah, it took me sometime to realize this merge from September last year isn't published. Scripts are blowing up because of it 😕

PR #1416, opened 2017-12-11
Doc parser: don't strip comment leader twice

mistydemeo:
This fixes a bug I encountered when upgrading a Hubot application with some scripts which used the deprecated documentation syntax. The comment leader is already stripped by the time Hubot hits the `if (currentSection === null)` loop, so the `.slice` strips the leading two characters which contain actual text. For example, here's how this line gets parsed:

```
#   hubot translate me <phrase> - The 'me' is optional
```

Before:

`bot translate me <phrase> - The 'me' is optional`

After:

`hubot translate me <phrase> - The 'me' is optional`

### DIFF
--- a/src/robot.js
+++ b/src/robot.js
@@ -524,7 +524,7 @@ class Robot {
     const body = fs.readFileSync(require.resolve(path), 'utf-8')
 
     const useStrictHeaderRegex = /^["']use strict['"];?\s+/
-    const lines = body.replace(useStrictHeaderRegex, '').split('\n')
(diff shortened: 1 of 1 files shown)


mistydemeo:
After looking into it a bit more, it looks like the actual problem is that somehow I ended up with a few scripts with CRLF separators; I'm not surprised Hubot choked on them, but stripping too many characters is an interesting way to treat it.

mistydemeo:
Updated to split on any newlines, not just `\n`; this fixes the issue with the scripts I had noticed. Supporting CRLF is a pretty niche usecase, but since these scripts worked fine in Hubot 2.x, I think it's worth fixing this as a regression.

Issue #1379, opened 2017-08-01
Express server and http client update


timkinnane:
Hi, can we bump the dependency for express to v4. It would need to be tested, but I think it's mostly compatible, just adds some features.

Also noted that [the http client used in hubot](technoweenie/node-scoped-http-client) is now marked as unmaintained. It does the job, so I'm not concerned, but thought maybe there's alternatives that will be better going forward?

elblivion:
(**bump**) We're also waiting for express v4 to fix a downstream vulnerable dependency, any chance this could be tackled?

PR #1425, opened 2018-02-15
Update Express to latest

elblivion:
See #1379

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "cline": "^0.8.2",
     "coffee-script": "1.6.3",
     "connect-multiparty": "^1.2.5",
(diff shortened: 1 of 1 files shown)


elblivion:
Awesome, I'd been sitting on this for a while... can't believe the timing. :-)

I was wondering how to test the HTTP endpoint, it's not actually tested. I'm a node noob but from skimming https://github.com/expressjs/express/wiki/Migrating-from-3.x-to-4.x it didn't look like it would be anything to do.

Will that change be merged here? What is currently the most maintained fork?

elblivion:
Hi, any chance of merging this or should I close it?

PR #1444, opened 2018-05-03
updated heroku deploy links

brntbeer:
👋 I'm going through documenting Hubot for some friends and coworkers and noticed some links not being correct (though the redirects work!).


brntbeer:
Changed my initial comment. I think this is fine to merge as is if someone else wants to give it the ol' 👍

brntbeer:
thanks @technicalpickles !!!

PR #1453, opened 2018-07-30
Bump express and connect-multiparty

mistydemeo:
These two dependencies were lagging behind the current published versions.

According to the [express migration guide](https://github.com/expressjs/express/wiki/Migrating-from-3.x-to-4.x), it doesn't look like Hubot hits any of the pain points there. The tests are also passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "chalk": "^1.0.0",
     "cline": "^0.8.2",
     "coffee-script": "1.6.3",
-    "connect-multiparty": "^1.2.5",
-    "express": "^3.21.2",
+    "connect-multiparty": "^2.1.1",
+    "express": "^4.16.3",
     "log": "1.4.0",
     "optparse": "1.0.4",
     "scoped-http-client": "0.11.0"


mistydemeo:
Node 4 failure looks unrelated.

mistydemeo:
Had the chance to test this with a production Hubot; it's responding to `GET` requests just fine.

technicalpickles:
Chatted with @mistydemeo about this one. There are express changes, so we are figuring this is a minor version bump.

Other problem with releasing this is that 3.x is still marked as `next` (https://github.com/hubotio/hubot/issues/1406). I'm not sure where things were left off blocking it (https://github.com/hubotio/hubot/issues/1057), but I think it has been out this long that we should tag this release as `latest` and sort out the problems that happen.

mistydemeo:
> There are express changes, so we are figuring this is a minor version bump.

👍 I'm figuring this should be 3.1.0 rather than 3.0.2.

mistydemeo:
Can confirm the Node 4 failure is occurring on master, and was not introduced in this branch: https://travis-ci.org/hubotio/hubot/builds/411422293

PR #1457, opened 2018-08-03
travis: update npm before running npx

mistydemeo:
Alternative to #1456.

npx is a part of npm itself as of 5.2.0, so this is probably the best way to make sure npx is available to us later in the script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,5 @@ deploy:
   provider: script
   skip_cleanup: true
   script:
+    - npm install -g npm@latest
     - npx travis-deploy-once "npx semantic-release"


technicalpickles:
I was curious if there's an official place to hook into Travis. Found this: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-a-specific-npm-version

```
before_install:
  - npm i -g npm@version-number
```

https://github.com/hubotio/hubot/pull/1456 failed without much insight. I was wondering if it had to do with doing an npm install during the deploy? At least if it's during `before_install`, all the builds should see it, so we get a failure during any build, rather than the only on one build when merged to master.

technicalpickles:
Got this fixed over in https://github.com/hubotio/hubot/pull/1458

PR #1472, opened 2018-10-16
Add jsonLimit setting that reads from ENV var

ibarreto:
This PR adds a `jsonLimit` setting so we can set the size for `express.json()`.


src/robot.js
```diff
@@ -441,7 +442,7 @@ class Robot {
     }
     app.use(express.query())
 
-    app.use(express.json())
+    app.use(express.json({ jsonLimit }))
```
ilyoff: according to documentation, there is no jsonLimit parameter, but limit

ilyoff:
```suggestion
    app.use(express.json({ limit: jsonLimit }))
```

ibarreto:
@ilyoff good catch, thanks

ibarreto:
This has been a very painful PR process, closing

PR #1474, opened 2018-10-18
Fix spelling mistake

sjetha-pason:
Reference for the correct spelling: https://en.wikipedia.org/wiki/Dr._Seuss


docs/scripting.md
```diff
@@ -94,7 +94,7 @@ the response object a private message can be sent to the original sender.
 ```coffeescript
   robot.respond /I don't like Sam-I-am/i, (res) ->
```
stephenyeargin: Can we also change the capitalization of `Sam-I-Am` in this section? It appears in several places below.

9780525579588_p0_v1_s550x406

sjetha-pason: Done.
I also lower-cased the matching text for "Sam-I-Am", since it's case insensitive.

PR #1482, opened 2018-11-27
Dependency updates

mistydemeo:
This makes a few changes to Hubot's dependencies:

* `coffee-script` has moved to `coffeescript`. The old name still works, but this silences an npm warning.
* Bump to the latest versions of both `mocha` and `coveralls`, both of which had lagged a bit behind upstream. Thankfully the tests run as-is without requiring changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "devDependencies": {
     "chai": "~2.1.0",
-    "coveralls": "^2.13.1",
+    "coveralls": "^3.0.2",
     "mocha": "^5.2.0",
     "mockery": "^1.4.0",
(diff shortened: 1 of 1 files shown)


technicalpickles:
https://github.com/hubotio/hubot/pull/1466/ looks like an earlier attempt at this

technicalpickles:
I ended up merging  https://github.com/hubotio/hubot/pull/1466 because it had some documentations update to go with coffeescript too. This PR is still useful since it addresses other dependencies.

mistydemeo:
Rebased and updated the commit names to be semantic-release compatible.

PR #1487, opened 2018-12-18
Brain: ensure User objects are actually User objects

mistydemeo:
While the brain consistently documents that its functions return `User` objects, this actually isn't guaranteed. If the data is serialized to JSON for persisting - [like the Redis brain does](https://github.com/hubotio/hubot-redis-brain/blob/master/src/redis-brain.js#L93) - then `User` objects are serialized as plain JS objects and parsed back as them too. This leads to a mix of plain JS objects and actual `User` instances. This can lead to confusion, and it's also causing problems with an in-development branch of mine which grants instance methods to `User` objects.

This PR resolves this by turning plain objects back into `User` objects in `mergeData` before calling `loaded`. The new objects retain all the properties of the original ones, ensuring they're consistent with any newly-constructed objects created at runtime.

cc @technicalpickles

### DIFF
--- a/src/brain.js
+++ b/src/brain.js
@@ -4,6 +4,28 @@ const EventEmitter = require('events').EventEmitter
 
 const User = require('./user')
 
+// If necessary, reconstructs a User object. Returns either:
+//
+// 1. If the original object was falsy, null
+// 2. If the original object was a User object, the original object
+// 3. If the original object was a plain JavaScript object, return
+//    a User object with all of the original object's properties.
+let reconstructUserIfNecessary = function (user) {
+  if (!user) {
+    return null
+  }
+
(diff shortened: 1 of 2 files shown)


mistydemeo:
> Are there any tests we can add to verify this behavior?

I'll look at adding tests.

> Can you think of any ways this would negatively impact existing hubot installs?

I can't think of any. Any existing installs would have to have been able to cope with mixed user types up until now, so having all user types coerced back into `User` wouldn't introduce a user type that those installs weren't already dealing with.

mistydemeo:
Updated to add tests; this confirms that `mergeData` with arbitrary non-User objects results in User objects being created.

mistydemeo:
Updated the commit message for semantic-release.

PR #1489, opened 2019-01-11
Add a new persistence model - Datastore

mistydemeo:
This pull request adds support for a secondary persistence model to complement the traditional Hubot brain.

My company's Hubot installation has been hitting some pain points with the traditional Hubot brain. The model of storing data in-memory, and persisting the entire structure on a schedule, has a few intractable pain points:

1. We run multiple Hubots, which want cooperative access to a database. This isn't feasible in a solution where the brain is persisted and loaded as a whole on a schedule.
2. Many scripts, including third-party scripts we have installed, access the underlying `robot.brain.data` structure directly instead of going via getters and setters. This is also true for user-scoped data, which currently doesn't provide any getter/setter model at all.
3. We can't feasibly subclass the existing brain because the documented behaviour of the brain's `get` and `set` methods is synchronous, whereas many database adapters (like the node mysql adapter) provide only asynchronous database access.
4. Data integrity depends on Hubot itself maintaining a valid in-memory structure. We've experienced periods of data loss when Hubot goes offline before a persistence event is executed, or when invalid data was inadvertently included in the brain.

I don't think Hubot's existing brain can be adapted to solve these concerns without breaking compatibility, and I think the brain is too central for breaking compatibility to be acceptable. As a result, I've created a new, optional datastore class which solves these problems by providing a more robustly persistent database. Like the brain, it stores key-value data where values can be arbitrary JSON-compatible structures. Its primary distinguishing features are:

1. Each fetch and write operation is backed directly by access to the underlying database. This permits multiple Hubot instances running on separate servers to cooperatively access the same data.
2. Because data is persisted on a per-write basis, the risk of data-loss from Hubot exceptions or downtime is reduced.
3. Data can be written to multiple namespaces; currently, this is defined to mean one global namespace along with per-user data.
4. Getters and setters are asynchronous and return promises; this makes it easier to interface with database adapters which only provide asynchronous data access.

This pull request introduces the abstract class and an in-memory datastore suitable for testing. I'm planning to open-source the MySQL datastore implementation I created for internal use. The documentation has also been updated to reflect the datastore's existence and the main differences between it and the brain.

cc @technicalpickles

### DIFF
--- a/docs/implementation.md
+++ b/docs/implementation.md
@@ -36,6 +36,8 @@ There are two primary entry points for middleware:
 
 ## Persistence
 
+### Brain
+
 Hubot has a memory exposed as the `robot.brain` object that can be used to store and retrieve data.
 Furthermore, Hubot scripts exist to enable persistence across Hubot restarts.
 `hubot-redis-brain` is such a script and uses a backend Redis server.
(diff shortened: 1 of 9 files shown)


mistydemeo:
I've open-sourced our MySQL-based datastore implementation: https://github.com/github/hubot-mysql-datastore I'm happy to transfer ownership of that into this organization or the `hubot-scripts` organization.

PR #1496, opened 2019-03-05
Brain: avoid retaining dangerous robot/brain references

mistydemeo:
It appears that Hubot is accidentally retaining some references to `robot` and `brain` in objects, leading to unserializable circular references. This fixes the issue by

a) Ensuring temporary assignment of `robot` to objects is removed, and
b) Tucking `robot` inside a getter function in `Brain`, instead of assigning it as a variable, like in the `User` object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "devDependencies": {
     "chai": "~2.1.0",
     "coveralls": "^3.0.2",
+    "is-circular": "^1.0.2",
     "mocha": "^5.2.0",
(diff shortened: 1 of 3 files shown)


mistydemeo:
> Looks good. Are there any tests you can think of that would need to be added?

Let me see if there's a decent npm module I can use to create tests for this. The problem is detecting the cyclic structure!

mistydemeo:
Updated to add tests using the `is-circular` module.

PR #1501, opened 2019-06-11
feat: Robot constructor to take an object of properties

pauloroberto2019:
Previously, Robot took positional arguments. This is fine for internal
use, but when programatically building, it looks weird because you have
a bunch of null and boolean values which don't make sense on their own.
It is bad for the reasons having multiple positional arguments is bad,
lack of readability.

Instead of positional arguments, this updates Robot to take an options
object, so the parameters are named and it is clear to a reader what
each does.

This continues to support the old positional arguments, but emits a
deprecation message with depd. Internal use of has been updated to use
the options object, and there probably aren't a ton of uses of Robot
directly in the wild (although there is definitely at least
hubot-test-helper). This support can be considered for removal during
the next breaking release.

This is part of https://github.com/hubotio/evolution/pull/3

### DIFF
--- a/bin/hubot.js
+++ b/bin/hubot.js
@@ -23,7 +23,7 @@ const options = {
   adapter: process.env.HUBOT_ADAPTER || 'shell',
   alias: process.env.HUBOT_ALIAS || false,
   create: process.env.HUBOT_CREATE || false,
-  enableHttpd: process.env.HUBOT_HTTPD || true,
+  httpd: process.env.HUBOT_HTTPD || true,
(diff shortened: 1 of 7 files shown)


pauloroberto2019:
package.json
src/robot.js

pauloroberto2019:
> package.json
> src/robot.js

pauloroberto2019:
package.json

chuotcongtu:
programaticly-configured-hubot-js

PR #1539, opened 2020-07-22
Update copyright date to 2020

iHiD:
This file is referenced in the [GitHub docs](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository) as an examplar licence. I figured it was worth therefore keeping it up to date :)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2017 GitHub Inc.
+Copyright (c) 2011-2020 GitHub Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


davemcg3 (changes requested):
Copyright notices are kind of a weird thing. There's an easily readable summary of the requirements here (assuming US domicile for github or hubotio): https://www.copyright.gov/circs/circ03.pdf

The short of it is that a copyright notice is entirely optional for any works published after March 1, 1989 and only need to consist of the copyright symbol, the word or abbreviation for copyright and the year of the first publication of the work (most people erroneously assume it should be the current year but by putting the year of first publication copyright extends from that year to the present). It would make more sense for your PR to remove the year you updated instead.

PR #1545, opened 2020-10-12
master

rajasibero:
https://gist.github.com/wjrogers/1016065#gistcomment-1954723


davemcg3 (commented):
These don't seem like helpful changes. You removed important text from the README and added a file of dubious merit specific to a single platform.

samuei (commented):
Hacktoberfest spam. Not intended to be useful.

PR #1557, opened 2021-07-20
threaded replies

esirko:
Building off of some earlier hackathon work to enable threaded replies in Slack messages for RPC chatops, I would like to extend it to all chatops. This is part of the changes that would be necessary to allow you to specify `--thread` or `--nothread` to any hubot chatop to make it threaded or not threaded.  

- https://github.com/github/hubot-slack/pull/21 for the prior work, for reference
- https://github.com/github/hubot-slack/pull/22 for a sibling PR to this one

### DIFF
--- a/src/listener.js
+++ b/src/listener.js
@@ -75,6 +75,27 @@ class Listener {
       // callback and calls done (never calls 'next')
       const executeListener = (context, done) => {
         this.robot.logger.debug(`Executing listener callback for Message '${message}'`)
+
+        // Thread style can be specified by the chatops script, and can be overridden by the user via requestedThreadStyle.
+        let threadStyle = 0
+        if (message.requestedThreadStyle !== undefined) {
+          threadStyle = message.requestedThreadStyle
+        } else if (this.options.threadStyle !== undefined) {
+          threadStyle = this.options.threadStyle
+        }
+
+        if (threadStyle > 0) {
+          if (context.response.message.rawMessage !== undefined) {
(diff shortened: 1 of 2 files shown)


joeyguerra:
I think threaded replies are better left to the chat adapter to implement instead of adding it as a core concept in Hubot.

adligit2 (approved):

adligit2 (commented):
review

src/listener.js
adligit2, line 1: submit

adligit2 (changes requested):
submit

adligit2:
```javascript

```

technicalpickles:
@adligit2 it looks like you may be testing something. Please find another place to do so.

joeyguerra:
My concerns with this PRs threaded reply design boils down to the fact that Slack's threaded replies concept leaks into Hubot's code. Refer to the companion code in [`github/hubot-slack`](https://github.com/github/hubot-slack/pull/22/files). It requires a property, `requestedThreadStyle` be set (responsibility) within Hubot's *core*.

An alternative design would be passing through an "adapter context" (a whole object instead of just a property) that can just be passed through to the adapter from within Hubot's *core*. That way, the only *concept* that Hubot is aware of is the concept of an "adapter context" which doesn't require Hubot to *know* about the details of that object. Hubot will just ensure that the object is passed through to the adapter. The Adapter can then use it to model a threaded reply.

PR #1576, opened 2022-12-17
Check for npm updates monthly

jlosito:
This will use dependabot to check for updates for any of the npm related packages this project uses and submit pull requests with version bumps in order to keep them up-to-date.


joeyguerra:
Dependabot has been enabled so I don't think we need this anymore.

PR #1579, opened 2023-03-23
Update LICENSE.md

TEE-EMM97:


joeyguerra (approved):

technicalpickles:
For reference, an previous PR to update this pointed out that it's not really needed to keep updating the year.

Issue #1546, opened 2020-10-13
Noticed this problem as well when trying to upgrade from 2.9.1 to 3.3.2, and using hubot-slack.  In 2.9.1 I had used robot.send to send a message, capture the returned array of promises to get a handle to the message, and then post additional messages in a thread under the first message.  Now this no longer works.


rajasibero:
Noticed this problem as well when trying to upgrade from 2.9.1 to 3.3.2, and using hubot-slack.  In 2.9.1 I had used robot.send to send a message, capture the returned array of promises to get a handle to the message, and then post additional messages in a thread under the first message.  Now this no longer works.
As xob explained, an easy fix here to restore the old behaviour would be to add the "return" statement in the robot.send function:
```
send (envelope/* , ...strings */) {
    const strings = [].slice.call(arguments, 1)

    return this.adapter.send.apply(this.adapter, [envelope].concat(strings))
  }
```
(and the same in robot.messageRoom)

_Originally posted by @vpeltola in https://github.com/hubotio/hubot/issue_comments/504529442_

joeyguerra:
Fixed with 3.4 release.

Issue #1530, opened 2020-06-06
Roadmap for Hubot


VinayaSathyanarayana:
What is the roadmap for Hubot?
I see last commit to Master branch in April 2019 (more than a year back)

joeyguerra:
We just released a new version to catch up. We should startup another discussion about the future.

Issue #1485, opened 2018-12-13
Remove node-scoped-http-client


sschueller:
The Library `node-scoped-http-client` is an unmaintained security risk. Last commit was 5 Jun 2015 and the maintainer is giving access to npmjs to anyone who is interested.

technicalpickles:
This was fixed in https://github.com/hubotio/hubot/pull/12

Issue #1483, opened 2018-11-29
Robot uses deprecated `require.extensions` prevents using Jest for testing


jeremygaither:
Node.js global `require.extensions` was deprecated since v0.10.6 and appears to have been removed from the documentation as of v8 LTS: 

This block of code causes problems when testing with Jest and [hubot-test-helper](https://github.com/mtsmfm/hubot-test-helper): https://github.com/hubotio/hubot/blob/ea182fa73852177897db51e119ce78e4c645bacc/src/robot.js#L352-L355

The module [hubot-test-helper](https://github.com/mtsmfm/hubot-test-helper) uses `const Hubot = require('hubot/es2015');` to require in Hubot: https://github.com/mtsmfm/hubot-test-helper/blob/master/src/index.js#L5

Global `require.extensions` is undefined running under the default Jest test runner (logging added to diagnose problem):

## failing scenario

```
  console.log node_modules/hubot-test-helper/src/index.js:161
    script /Users/REDACTED/scripts/REDACTED.js

  console.log node_modules/hubot/src/robot.js:351
    filepath /Users/REDACTED/scripts filename REDACTED.js

  console.log node_modules/hubot/src/robot.js:354
    full /Users/REDACTED/scripts/REDACTED ext .js

  console.log node_modules/hubot/src/robot.js:356
    in robot require undefined

  console.log node_modules/hubot/src/robot.js:361
    not require extensions
```

Related: https://github.com/facebook/jest/issues/281 (workarounds are not working for Hubot)

Is there another recommended way to use Jest with Hubot, or is the preferred/recommended testing framework Mocha?

Would it be better to remove this check that is relying on a deprecated object to exist? The code does indeed run in Jest with the default runner and minimal config changed when the above block is removed. I don't know if https://github.com/hubotio/hubot/issues/1355 is still a concern. It does attempt to load non-js files (and folders that may not have an index.js) but the try/catch block can avoid problems if `process.exit(1)` is removed.

## working modified scenario

```
    console.log node_modules/hubot-test-helper/src/index.js:161
      script ../scripts
    console.log node_modules/hubot/src/robot.js:351
      filepath /Users/REDACTED/scripts filename deploy-queue.js
    console.log node_modules/hubot/src/robot.js:354
      full /Users/REDACTED/scripts/deploy-queue ext .js
    console.log node_modules/hubot/src/robot.js:356
      in robot require undefined
    console.log node_modules/hubot/src/robot.js:369
[...]

PR #1581, opened 2023-04-13
2023 catchup

joeyguerra:
# Existing PRs and Issues
- fix(hubot): Can we please get rid of the SIGTERM catch? #1403
- fix(robot): Set max json request body size with EXPRESS_LIMIT #1405
- fix(robot): express basicAuth #1470
- fix(robot): Robot uses deprecated require.extensions prevents using Jest for testing #1483 
- fix(http): Remove node-scoped-http-client #1485
- docs(windows): Fix setx typo to set and put double quotes in the correct place #1509
- docs(roadmap): Roadmap for Hubot #1530
- fix(robot): return results from `send` and `messageRoom` #1546 
- fix(http): HUBOT_HTTPD can now disable httpd #1548
- docs(patterns): use scoped npm package as adapter #1564

# New from [hubot-new/hubot](https://github.com/hubot-new/hubot) fork

- fix(package): Add package-lock.json generated with Node v14 / NPM v6 [#32](https://github.com/hubot-new/hubot/issues/32)
- fix(shell): Every other character is skipped when typing at the prompt with the default Shell adapter in Node.js versions > 16.19.0 [#39](https://github.com/hubot-new/hubot/issues/39)
- fix(actions): GitHub Actions for tests [#45](https://github.com/hubot-new/hubot/issues/45)
- chore(deps): bump lodash from 4.17.11 to 4.17.21 [#47](https://github.com/hubot-new/hubot/pull/47)
- chore(deps): bump minimatch and mocha [#48](https://github.com/hubot-new/hubot/pull/48)
- chore(deps): bump semver-regex and semantic-release [#49](https://github.com/hubot-new/hubot/pull/49)
- chore(deps): bump shelljs and standard [#51](https://github.com/hubot-new/hubot/pull/51)
- chore(deps): bump path-parse from 1.0.6 to 1.0.7 [#54](https://github.com/hubot-new/hubot/pull/54)
- chore(deps): bump hosted-git-info from 2.7.1 to 2.8.9 [#55](https://github.com/hubot-new/hubot/pull/55)
- chore(deps): bump handlebars from 4.1.0 to 4.7.7 [#56](https://github.com/hubot-new/hubot/pull/56)
- chore(deps): bump y18n from 4.0.0 to 4.0.3 [#57](https://github.com/hubot-new/hubot/pull/57)
- chore(deps): bump yargs-parser and nyc [#58](https://github.com/hubot-new/hubot/pull/58)
- chore(deps): bump express-basic-auth from 1.1.5 to 1.1.7 [#59](https://github.com/hubot-new/hubot/pull/59)
- fix(tests): Include CI for Windows, OSX [#62](https://github.com/hubot-new/hubot/pull/62)

### DIFF
--- a/.github/workflows/nodejs-macos.yml
+++ b/.github/workflows/nodejs-macos.yml
@@ -1,0 +1,26 @@
+name: Node.js (macOS) CI
+
+on:
+  push:
+    branches: [ "master" ]
+  schedule:
+    - cron:  '5 4 * * 0'
+
+jobs:
+  npm-test:
+
+    runs-on: macos-latest
+
+    strategy:
+      matrix:
+        node-version: [18.x]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: npm test
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,37 +1,32 @@
 # Hubot Roadmap
 
-Hubot v3 aims to be a bot framework optimized for developers and developer workflows, with great integration with the most popular chat clients and developer tools, and an active community that is sharing scripts and best practices.
+Hubot aims to be a bot framework optimized for developers and developer workflows, with great integration with the most popular chat clients and developer tools, and an active community that is sharing scripts and best practices.
 
-This roadmap represents some of priorities for us over the next couple months. Issues or pull requests will be opened to discuss each of these items as they progress.
+This roadmap represents some priorities for us over the next couple months. Issues or pull requests will be opened to discuss each of these items as they progress.
 
 ## 1. Return to a “maintained” status
 
-- [x] Create a Hubot core team with at least 2 GitHub employees and at least 1 community member. The core team is [@technicalpickles](https://github.com/technicalpickles), [@bkeepers](https://github.com/bkeepers), [@mose](https://github.com/mose), and [@gr2m](https://github.com/gr2m) ([#1323](https://github.com/github/hubot/pull/1323))
+- [x] Create a Hubot core team with invested users with no direct connection, interested in continuing the #hubot project.
+- [x] Review all [open PRs](https://github.com/github/hubot/pulls) and triage [open issues](https://github.com/github/hubot/issues)
 - [ ] Document all maintainer processes (triage, release, etc.)
 - [ ] Create an issue template that addresses common requests
-- [ ] Configure automation ([probot stale](https://github.com/probot/stale), [Greenkeeper](https://greenkeeper.io/), [semantic-release](https://github.com/semantic-release/semantic-release))
-- [ ] Review all [open PRs](https://github.com/github/hubot/pulls) and triage [open issues](https://github.com/github/hubot/issues)
-- [ ] Establish a release process and regular release cadence of the first Tuesday of every month.
-- [ ] Establish a first-responder rotation, which will aim to reduce the average time to first response on all new Issues and PRs to 48 hours.
+- [ ] Establish a release process and regular release cadence.
 
 ## 2. Modernize the community
 
-- [ ] Consolidate all officially supported Hubot projects into a single GitHub organization. This will include github/hubot and a handful of supported scripts, but will not include all community scripts in https://github.com/hubot-scripts ([#1327](https://github.com/github/hubot/issues/1327))
-- [ ] Create a community forum to provide a place for people to ask questions, get help, and share best practices. [Discourse](https://www.discourse.org/) is the obvious choice here.
-- [x] Choose a chat platform for maintainers and contributors, and post notices in various existing places (#hubot on freenode, github/hubot on Gitter). Slack is the obvious choice here. [Join us on Slack](https://hubot-slackin.herokuapp.com/).
-- [x] Add a code of conduct based on http://contributor-covenant.org/ and processes to enforce it in all official spaces. ([#1334](https://github.com/github/hubot/pull/1334))
-- [ ] Publish weekly community updates (blog, newsletter, etc) which highlight recent and upcoming changes, give shoutouts to contributors / maintainers, and maybe mention interesting uses of Hubot
 - [x] Create Hubot Evolution—inspired by [Swift Evolution](https://github.com/apple/swift-evolution)—for proposing user-visible enhancements. ([hubotio/evolution#1](https://github.com/hubotio/evolution/pull/1))
+- [ ] Review and update the [contributing guide](CONTRIBUTING.md)
+- [ ] Review hubot scripts and discuss whether or not we should fork them into this org or not.
+- [ ] Do we want a chat community for Hubot?
+- [ ] Review the [code of conduct](CODE_OF_CONDUCT.md) based on http://contributor-covenant.org/
+- [ ] Decide on setting up Github Pages
 
 ## 3. Modernize the project
 
-Each of these proposals will go through the [Hubot Evolution](https://github.com/hubotio/evolution) process.
-
-- Translate from CoffeeScript to JavaScript and update to modern versions of Node.js and NPM (or Yarn)
+- Update to modern versions of Node.js
 - Revisit new bot generator (yeoman has a ton of dependencies, some of which can be error prone on windows)
 - Support for running multiple adapters and archetypes (chat, deployment, CI, github, etc)
-- Merge with [@probot](https://github.com/probot) and build out first class GitHub integration.
-- Introduce "Commands”, an explicit interface for registering commands (like Slack’s slash commands) as an alternative to regular expressions
-- Publish a ChatOps RPC spec and implement support for Hubot acting as both a client and a server.
+- Discuss introducing the `Commands` concept as an interface for registering commands (like Slack’s slash commands) as an alternative to regular expressions
+- Discuss publishing a ChatOps RPC spec and implement support for Hubot acting as both a client and a server.
 - Support rich messages and interactions on platforms that support it
 - Publish a public script directory backed by NPM
--- a/src/httpclient.js
+++ b/src/httpclient.js
@@ -1,0 +1,282 @@
+/*
+Copyright (c) 2014 rick
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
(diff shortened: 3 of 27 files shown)


xurizaemon:
@technicalpickles @joeyguerra I think we should see Github Actions kick in on this MR - does that need to be enabled on the repo?

Is there a task also to remove the Travis CI checks as well, if they're sufficiently replaced by Github Actions here? They didn't show up on hubot-new/hubot but we wouldn't have had the expected configuration in place.

xurizaemon:
https://github.com/hubotio/hubot/pull/1581/commits/3ceeb0a4881270ecc9d134e25aaf1806f0236526 may want to be revisited - that change can be updated to retain the existing org.

xurizaemon:
Whooo, this PR adds 10K LoC. Is that all from .coffee > .js?!

joeyguerra:
It's the `package-lock.json` file. 9,494 LOC.

src/robot.js
```diff
@@ -352,7 +352,7 @@ class Robot {
     const full = path.join(filepath, path.basename(filename, ext))
 
     // see https://github.com/hubotio/hubot/issues/1355
-    if (!require.extensions[ext]) { // eslint-disable-line
+    if (['.js', '.mjs'].indexOf(ext) == -1) { // eslint-disable-line
```
technicalpickles, line 355: Does this have the effect of removing support for coffescript? I know there will be coffeescript based scripts out there both open source and private that are kicking around.

The reason we used this originally was that coffeescript and (I think) typescript would add to this. Looks like officially it's deprecated:

> In the past, this list has been used to load non-JavaScript modules into Node.js by compiling them on-demand. However, in practice, there are much better ways to do this, such as loading modules via some other Node.js program, or compiling them to JavaScript ahead of time.
>
>Avoid using require.extensions. Use could cause subtle bugs and resolving the extensions gets slower with each registered extension.

I think we need this for compatability for now, until there is a way to bundle your bot à la webpack or something similar. So, if we can disable eslint on this line, that would be preferred.

src/httpclient.js
```diff
@@ -0,0 +1,282 @@
+/*
```
technicalpickles, line 1: Can we document why this is included? I assume it has to do with not getting updates to scoped-http-client.

.github/workflows/nodejs-macos.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  push:
+    branches: [ "master" ]
+  schedule:
+    - cron:  '5 4 * * 0'
+
+jobs:
+  npm-test:
+
+    runs-on: macos-latest
```
technicalpickles, line 12: Not a blocker, but is it possible to use the same workflow for all 3 platforms? Or is it to limit testing the CI to only run the latest node on macOS and windows, and let ubuntu test the spread of supported versions?

The matrix strategy in nodejs-ubuntu.yml supports specifying os too.

ROADMAP.md
```diff
@@ -1,37 +1,32 @@
 # Hubot Roadmap
```
technicalpickles, line 1: I think it can be safe to remove this file to reflect there isn't one currently 🙊

technicalpickles (commented):
Awesome, thank you for taking the time to put this together as a PR 🚀 

Overall, I see no issue. The thing that gives me pause is around detecting which javascript files to load. As it is, I _think_ it would force everyone to migrate their scripts to coffeescript to javascript if they haven't already. They really ought to be, but it's a question if we want to force this now or in a followup. What do y'all think?

technicalpickles:
> does that need to be enabled on the repo?

It is already enabled 🤔  They do require an initial approval for first time contributors, but I am not seeing a way to enable them. Will see if I can figure it out.

technicalpickles:
I have no idea why I can't approve or run GitHub Actions for this branch (followed the [docs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) and there's not an approve workflow as suggested).

I was able to `npm install` and `npm test` locally on macOS w/ Node 18.x. Two caveats though.

1) npm install had some warnings about an old lockfile. May want to `npm install` and commit the updated package.json at some point (doesn't need to be now)

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

2) `npm test` is successful, but seems to hang after printing out the summary:

```

load: 7.47  cmd: node 44076 waiting 1.04u 0.12s
load: 7.47  cmd: node 44076 waiting 1.04u 0.12s
load: 7.47  cmd: node 44076 waiting 1.04u 0.12s
^C----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------|---------|----------|---------|---------|-------------------
All files |       0 |        0 |       0 |       0 |
----------|---------|----------|---------|---------|-------------------
```

Looks like it may be `nyc` which is doing... coverage reporting? I'm not very worried about this, but should be fixed up as a followup.

src/robot.js
joeyguerra: What's wrong with just adding `.coffeescript` to the array?

xurizaemon:
That "pending" thing was raised in https://github.com/hubot-new/hubot/issues/33 which is closed but IDK if the issue was resolved.

Is it that Travis failing prevents Github Actions from proceeding perhaps?

src/httpclient.js
joeyguerra: Yes. Do you have a suggestion where to document that? Yes, the issue was scoped-http-client wasn't maintained anymore.

ROADMAP.md
joeyguerra: Sounds good.

src/robot.js
joeyguerra: I've added `.coffee` to the array and it loaded a `.coffee` script.

.github/workflows/nodejs-macos.yml
joeyguerra: Can we change this in a subsequent PR? Adding testing multiple node versions on all OSs does increase the pipeline time.

ROADMAP.md
joeyguerra: I've removed it. I'm guessing we'll want to create a new one.

technicalpickles:
> Is it that Travis failing prevents Github Actions from proceeding perhaps?

Doubtful. Looking more closely, it's a branch protection for the default branch, and 'waiting for status to be reported' suggests to me it's not getting kicked off at all. I can remove that requirement for now.

technicalpickles:
Something I'd like to double check before approving & merging is trying to upgrade an existing hubot. https://github.com/hubotio/hubot-for-hubot is one we can use to test. I'm more worried about loading and running the tests than anything else, so testing with the shell adapter is fine instead of getting it hooked back up to the hubotio slack.

.github/workflows/nodejs-macos.yml
xurizaemon: https://github.com/hubot-new/hubot/pull/62#issuecomment-1500961898 was the reasoning, no objection to moving that to a common file if Actions will support that syntax. Making all PRs run on Windows would add a few minutes to each pipeline execution, running Ubuntu on PRs was around the 30s mark.

- Windows - 1m45 @ https://github.com/hubot-new/hubot/actions/runs/4648747554
- Ubuntu - 48s @ https://github.com/hubot-new/hubot/actions/runs/4648748957
- MacOS - 54s @ https://github.com/hubot-new/hubot/actions/runs/4648746939

Hardly a killer wait, but enough that when I turned it on, I didn't want to double (or more, with a matrix) test run time for what I felt was likely not a huge increase in coverage. There was one code location (hubot-new/hubot@3743c7fc8ecb709422290f50837f685d0990af7a) where OS-specific code was affected, but we're also validating that dependencies work on various OS.

xurizaemon: Will leave this PR as-is for now.

I've opened https://github.com/hubotio/hubot/issues/1582 to document this change in follow-up and propose removing the TravisCI configuration. (I don't have anything against TravisCI, but it's one more thing to maintain and Github Actions does what we need IMO.)

I agree that a single workflow configuration would be clearer; if we _have_ to use multiple then the yaml should have a comment explaining why.

xurizaemon:
@joeyguerra don't forget the commit message syntax in https://github.com/hubotio/hubot/blob/master/CONTRIBUTING.md for semantic-release. (I'll eventually add a workflow to ensure that, but not here.)

joeyguerra:
@xurizaemon any suggestions on what to put for the title since this PR includes a mixture of multiple fixes, features, doc updates?

xurizaemon:
Uhmm ... C-C-C-C-C-C-COMBO? Not sure about the issue title sorry.

AFAIK it's commit messages that semantic-release would care about, and it's likely to use the individual branch commits rather than the merge commits. Apologies if I didn't make my thinking explicit enough when [suggesting this](https://toot.cafe/@xurizaemon/110187582549891369) a couple days ago!

joeyguerra:
Ah. Now I see what you're saying. I was thinking, "surely we don't have to go back and update every commit message". But, in fact, you're saying we do. But those individual commits were not "atomic" in that they encapsulated a complete thought or change. Any suggestions for that situation?

technicalpickles:
The upside of travis not working is it won't be enforced. I can also bypass it as a maintainer even if it did.

As a followup, using PR titles instead of commits would be more practical and easier to manage instead of each commit.

src/httpclient.js
technicalpickles: I was thinking a comment at the top of the file. Would be good to move off this, but will be a pain because of how much plugins depend on it 😒

src/robot.js
technicalpickles: It's more that it's a hard-coded list, and there could be other extension types people register outside the context of hubot.

Based on my reading of the docs, the goal should be to transpile to javascript instead, so it becomes moot.

technicalpickles (commented):
Gave this a test drive on hubot-for-hubot locally. The local coffeescripts worked fine. I was having problems using external-scripts.json though, but that seems to be more about using `npm link` or `npm install ../hubot` than the PR.

I think we can call this good! Is there anything else to fix/update before merging? Can always do it in followups too.

>  mixture of multiple fixes, features, doc updates?

I'd lean towards feature, ie minor version.

joeyguerra:
When testing with hubot-for-hubot, http client bugs were exposed. buildOptions (extend) isn’t working as expected. I pushed a fix to this PRs branch. I’ll test it again with hubot-for-hubot later this evening.

joeyguerra:
I'm done with my testing as well. I pushed some bug fixes to the `httpclient` module after testing with hubot-for-hubot. @technicalpickles good call! Thank you.

@technicalpickles do you know how to do a dry run semantic-release so we can see how the commit messages are going to look? I was wondering if we should create a different branch, put all the changes that are in this PR into that new one and start from there instead to shore up the commit messages, but I don't know how semantic-release builds the change log from the commit messages. The documentation says it parses the commit messages between the previous release (git tag) and the new one. But I'm still unclear what commit messages it's going to get.

technicalpickles:
> do you know how to do a dry run semantic-release so we can see how the commit messages are going to look?

I'm not super familiar, as I wasn't the main proponent of using this. I pulled it down locally, and looks like there is a `--dry-run` command at least. It looks like it needs some GitHub and/or NPM tokens to run.

>  The documentation says it parses the commit messages between the previous release (git tag) and the new one. But I'm still unclear what commit messages it's going to get.

My hunch is that it wouldn't even run, because it was part of Travis CI which isn't running.

joeyguerra:
Ok. I'm ready whenever you are :) What are the next steps?

xurizaemon:
FWIW - I proposed (https://github.com/hubot-new/hubot/issues/44) we adopt semantic-release pattern based on seeing it documented in CONTRIBUTING.md here.

I supported that approach because it's easier to have a clear process and adopting an existing established process would make more sense for community adoption than change. It seems low-maintenance which I think is great. But it's a little opaque - I tried a few combos just now and couldn't work out how to get it to simulate generating release notes.

OTOH if there's a better method, we're free to do that too! So this comment is mostly to say we can be flexible on it :)

technicalpickles:
> when you say, "doing the release manually", do you mean execute the semantic-release cli command from your local machine to publish an updated package to NPM? Or adding the semantic-release command to one of (or create a new one) the GitHub Actions workflow yaml file?

I mean publishing with the npm command, and tagging git. I can create new "Release" really easily, because it can figure most of that out automatically nowadays (that didn't exist when semantic-release was being used)

technicalpickles:
> Ok. I'm ready whenever you are :) What are the next steps?

I'm going to:

- merge
- add y'all as maintainers
- see if I can release a one-off w/o using semantic release

technicalpickles (approved):

joeyguerra:
I'm trying a dry run with `semantic-release` and running into errors. First one is "can't login with password" :(

Issue #1082, opened 2015-11-12
Document writing scripts in Javascript (and other languages)


technicalpickles:
Hubot scripts can be written in Javascript in addition to Coffeescript. Hubot can actually load anything that nodejs can `require`. This isn't written down anywhere though.

cc @dev0n

Dev0n:
Thanks @technicalpickles. It would be nice to load directories as well as single script files.

For example:

``` scripts/foo ``` (providing an index.js is present inside foo).

cycomachead:
> For example:
> 
> scripts/foo (providing an index.js is present inside foo).

Yeah, this would be great as an addition, though it can also be done "manually".

I have this script which individually `require`s all scripts in subdirectories, but you could fairly easily modify it to use use `require(dir)`.

I'm fairly certain this script could just be turned into a package, as long as you can reliably get the location of the `scripts/` directory. (Which is be possible, but I haven't put much though into how brittle it would be.)

technicalpickles:
This will be done as part of moving to Javascript cc https://github.com/hubotio/evolution/pull/4 @gr2m

PR #1587, opened 2023-04-17
docs:  #1082 Update docs from CofeeScript to JavaScript examples

joeyguerra:
Addresses issue #1082 . Update the documentation to use JavaScript examples instead of CofeeScript.


technicalpickles:
An idea for maybe later... I wonder if we can run the code examples in this as part of CI? 🤔

joeyguerra:
@technicalpickles I added you to the reviewers list because I assume that's required to pass the "review required" requirement.

docs/patterns.md
```diff
@@ -152,54 +167,69 @@ One of the awesome features of Hubot is its ability to make changes to a product
...
+
 * Selectively allowing listeners (all listeners are power commands, except for listeners A, B, C, which are considered normal commands) is a more secure method (new listeners are restricted by default), but has a much higher maintenance overhead (every silly/fun listener needs to be explicity downgraded to "normal" status).
-* Selectively restricting listeners (listeners A, B, C are power commands, everything else is a normal command) is a less secure method (new listeners are put into the normal category by default, which could give unexpected access; external scripts are particularly scary here), but has a lower maintenance overhead (no need to modify/enumerate all the fun/culture scripts in your access policy).
+* Selectively restricting listeners (listeners A, B, C are power commands, everything else is a normal command) is a less secure method (new listeners are put into the normal category by default, which could give unexpected access; external scripts are particularly dangerous here), but has a lower maintenance overhead (no need to modify/enumerate all the fun/culture scripts in your access policy).
```
technicalpickles, line 188: I think "risky" is probably a better word. Dangerous implies not to use it, but risk suggests that it is something to consider before using.

docs/scripting.md
```diff
@@ -4,45 +4,49 @@ permalink: /docs/scripting/
...
-module.exports = (robot) ->
-  # your code here
+```javascript
+module.exports = (robot) => {
+  // your code here
+}
 ```
 
-The `robot` parameter is an instance of your robot friend. At this point, we can start scripting up some awesomeness.
+The `robot` argument is an instance of your robot friend. At this point, we can start scripting up some awesomeness.
```
technicalpickles, line 25: I think technically Javascript refers to them as parameters, but they are effectively synonyms.

docs/scripting.md
```diff
@@ -52,239 +56,252 @@ The `robot.respond /open the pod bay doors/i` callback is only called for messag
 It wouldn't be called for:
 
 * HAL: please open the pod bay doors
-   *  because its `respond` is bound to the text immediately following the robot name
+   *  because its `respond` is expecting the text to immediately follow the robot name
```
technicalpickles, line 59: I think "bound" is used because it's using a embedding the parameter passed into a regex with robot's name at the beginning.

Maybe a clearer way? "respond is expecting the text to be prefixed with the robot's name"?

joeyguerra:
Thank you for reviewing this so thoroughly. I've replaced `response` with `res`, argument(s) with parameter(s) because it makes sense to me as well. I left the `help.js` file because the documentation references it. But I don't have any other reason for including it.

scripts/help.js
```diff
@@ -0,0 +1,20 @@
+// Description:
```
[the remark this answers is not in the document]
technicalpickles: Ah, I think that is actually referring to the help script that would have been auto-generated. I'm not sure when it changed, but that's actually an external script now: https://github.com/hubotio/hubot-help

We do generate an example.js with a bunch of examples though: https://github.com/hubotio/generator-hubot/blob/master/generators/app/templates/scripts/example.js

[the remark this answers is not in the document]
joeyguerra: Are you saying I should remove it then?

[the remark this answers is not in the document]
joeyguerra: If it was loaded on a default instance of Hubot, I could see it being useful. But it doesn't, so it doesn't "work" on default. I'm going to remove it.

docs/patterns.md
```diff
@@ -152,54 +167,69 @@ One of the awesome features of Hubot is its ability to make changes to a product
 
 There are a variety of different patterns for restricting access that you can follow depending on your specific needs:
 
-* Two buckets of access: full and restricted with whitelist/blacklist
+* Two buckets of access: full and restricted with include/exclude list
 * Specific access rules for every command (Role-based Access Control)
-* Blacklisting/whitelisting commands in specific rooms
+* Include/exclude listing commands in specific rooms
```
technicalpickles, line 172: Just wanted to call this out, thanks for making the change! It is a change I try to do whenever I notice it.

technicalpickles (approved):

PR #1608, opened 2023-04-21
docs(development): Sample adater code wasn't working

joeyguerra:
#1609


technicalpickles (approved):
This definitely would be a good one to scrap docs and run during CI!

Issue #1611, opened 2023-04-21
`async` module is VERY out of date


joeyguerra:
The `async` module is very out of date. There's a PR (#1565 ) to update it, but that's out of date too. Latest version of `async` is 3.2.4 as of today.

PR #1565, opened 2022-04-18
chore(package): bump async

Torgen:
Of the version that fix CVE-2021-43138, tests passed with 2.6.4 and failed with 3.2.2.

PR #1612, opened 2023-04-22
chore(package): bump async

joeyguerra:
Reference PR #1565 .

Bumping `async` version to 2.6.4. I tried the latest version, 3.2.4 but tests were failing. So I'll have to create a different PR for that update since it could be more drastic.

Fixes issue #1611


joeyguerra:
I was able to fix the failed tests when bumping to version 3.x #1613

Issue #1605, opened 2023-04-21
fix deprecated dependencies


joeyguerra:
Hubot is old. There are some dependencies that are deprecated, no longer maintained. The following is the output from NPM with Node.js v18.3.0. The current Node.js LTS version is v18.16.0.

I'm guessing that in order to resolve these issues, we might have to get rid of some of the dev dependencies. For instance, I just noticed that `coveralls` depends on a version of `request` that is out of date and has a [security issue](https://github.com/hubotio/hubot/security/dependabot/4).

```bash
➜  hubot git:(master) ✗ npm i
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated formatio@1.1.1: This package is unmaintained. Use @sinonjs/formatio instead
npm WARN deprecated samsam@1.1.2: This package has been deprecated in favour of @sinonjs/samsam
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

added 866 packages, and audited 1103 packages in 5s

175 packages are looking for funding
  run `npm fund` for details

2 moderate severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.

Run `npm audit` for details.```

technicalpickles:
I renamed this "fix deprecated dependencies", which I think reflects the work. "Modernize the code base" can mean a lot more.

coveralls is a code coverage tool, with a service backing it. If it's not useful, then can drop it if it's not providing value AND has deprecations.

joeyguerra:
I don't yet have a scenario where I would get value from coveralls. I could see using it if we were strictly using TDD as our designing process. But I don't want to be dogmatic about TDD. I prefer balancing tradeoffs instead of apply a general rule across the board.

I'll create a PR to remove coveralls.

joeyguerra:
Dependencies have been updated to their latest versions. We'll work on replacing the log module in a future release.

PR #1613, opened 2023-04-22
chore(package): bump async to version 3

joeyguerra:
# Intent

#1605  Update dependencies to the latest versions

# Approach

The tests were timing out when updating `async` to v3.2.4. After investigating, the problem was that the `async.detectSeries` callback was not executing unless `message.done` was set to `true` in a Hubot Listener.

This [SO](https://stackoverflow.com/questions/42248993/how-to-return-output-in-node-using-async-detect-series) question and answer lead me to the issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0-development",
       "license": "MIT",
       "dependencies": {
-        "async": ">=0.1.0 <1.0.0",
+        "async": "^3.2.4",
         "chalk": "^1.0.0",
         "cline": "^0.8.2",
         "coffeescript": "1.6.3",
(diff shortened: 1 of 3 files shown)


joeyguerra:
@xurizaemon @levenleven @technicalpickles what do ya'll think about this PR?

xurizaemon:
Change looks fine. Please amend commit message (`asyn` => `async`), then good to go!

xurizaemon (changes requested):
Change looks fine. Please amend commit message (`asyn` => `async`), then good to go!

xurizaemon (changes requested):
Hey @joeyguerra 

The PR title has been corrected but the commit still says `asyn` for me. We need to be fixing the _commit messages_.

image

`git commit --amend` is the command to change the most recent commit.

joeyguerra:
It feels so wrong to me to `git push --force`.

xurizaemon (approved):

xurizaemon:
Heh :) yeah it smells bad, but it's better than the alternatives of a new PR for each change, or incorrect history entries, so it's the least bad correct action?

PR #1617, opened 2023-04-23
chore(package): Updating and removing dependencies

joeyguerra:
I removed `chalk` since it was only used to make some text bold in the Shell Adapter. I don't think bolding was worth including all the `chalk` dependency surface area.

I also added tests for the Shell adapter since this touches that file.

#1605

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MIT",
       "dependencies": {
         "async": ">=0.1.0 <1.0.0",
-        "chalk": "^1.0.0",
         "cline": "^0.8.2",
         "coffeescript": "1.6.3",
         "connect-multiparty": "^2.1.1",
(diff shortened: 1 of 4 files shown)


technicalpickles:
> I don't think bolding was worth including all the chalk dependency surface area.

The main reason for bold text is to make the shell more usable, because otherwise the text looks the same. I could see a case for choosing more than just bold.

Moving the adapter to its own repo could be another possibility.

joeyguerra:
I think there's value in keeping the Shell in here and the default adapter so that I can use it right off the bat. But I'd rather have less dependencies the bold text for a scenario that is probably short term. As in I'm only using the default Shell adapter for a few minutes before wanting to either build my own, or use a different adapter.

joeyguerra:
Btw, I was thinking that after merging this request, we could do another release. And then move onto updating the log module.

technicalpickles:
> But I'd rather have less dependencies the bold text for a scenario that is probably short term. As in I'm only using the default Shell adapter for a few minutes before wanting to either build my own, or use a different adapter.

For what it is worth, I've always used the shell adapter when building out scripts initially, so it wasn't ever a 'short term' thing.

If you want to drop the dependency, I can understand that. I'd recommend coding our own bold then, it's just a matter of outputting the right escape codes, and then resetting them after. You could even look at the chalk implementation.

joeyguerra:
> If you want to drop the dependency, I can understand that. I'd recommend coding our own bold then, it's just a matter of outputting the right escape codes, and then resetting them after. You could even look at the chalk implementation.

Done.

technicalpickles (approved):

technicalpickles:
> Btw, I was thinking that after merging this request, we could do another release. And then move onto updating the log module.

Could you make an issue to track that? Just don't want to lose track. We should get semantic-release going too

joeyguerra:
Yes. #1620

PR #1616, opened 2023-04-22
feat(log): bump log module from 1.4.0 to 6.3.1

joeyguerra:
# Intent

Update to the latest version

# Approach

Sinon couldn't wrap the `warning` function in the `log` module when I updated `log` to the latest version. So I created an object to decorate the underlying `log` module.

This means that technically the `logger` API has changed.

# Reasoning

I didn't want to change the tests because, to me, they represent the agreed upon public API. So my thought is that if the tests pass, the probability of breaking a dependent is low.

#1492

Issue #1492, opened 2019-02-01
Massive logger messages


terorie:
<img width="579" alt="image" src="https://user-images.githubusercontent.com/21371810/52141181-2c5ed980-2655-11e9-86a6-bfcb5668a449.png">

`[Fri Feb 01 2019 19:09:59 GMT+0100 (Central European Standard Time)]` is really hard to read as a prefix.

How about `[2019-02-01 19:09:59 GMT+1]`?

joeyguerra:
I think `new Date` is hard coded so you'd have to change the `Log` module.
```log: function(levelStr, args) {
    if (exports[levelStr] <= this.level) {
      var msg = fmt.apply(null, args);
      this.stream.write(
          '[' + new Date + ']'
        + ' ' + levelStr
        + ' ' + msg
        + '\n'
      );
    }
  },```

PR #1618, opened 2023-04-23
feat(log): Replace log module with the pino module

joeyguerra:
@levenleven suggested replacing `log` with `pino`. So here's a PR so we can test it, compare to `log` and discuss.

#1616 #1492

### DIFF
--- a/.github/workflows/nodejs-ubuntu.yml
+++ b/.github/workflows/nodejs-ubuntu.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, latest]
+        node-version: [14.x, 16.x, 18.x, latest]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


technicalpickles:
Is `log` deprecated? I didn't see it on the list on https://github.com/hubotio/hubot/issues/1605

joeyguerra:
I didn't see a deprecation warning, but it's been a while since someone made a commit to the repo: https://github.com/medikoo/log.

I created separate a separate PR for updating `log` because it's a significant change. So I'm treating it as part of a feature release, separate from this upcoming one with just updating dependencies.

This particular PR (using `Pino` instead) is just to serve as a place to discuss either continue using the `log` module, or replace it with `Pino` so that we can have structured logging.

joeyguerra:
@technicalpickles @xurizaemon @levenleven Ok ya'll. This PR replaces the logging module with Pino which has structured/ndjson logging output. Please review.

Note that I believe this is a major version bump. Let me know if you think otherwise.

joh6nn:
I would agree that this is a major bump

joeyguerra:
Just needs someone to approve.

joeyguerra:
@xurizaemon do you have a few minutes to review this PR?

technicalpickles (approved):

PR #1634, opened 2023-05-31
feat: load adapter from local path with .mjs file extension; load .mjs scripts

joeyguerra:
BREAKING CHANGE: Loading the adapter is done with async/await semantics. You can now specify an adapter with the .mjs file extension and Hubot will look in ./src/adapters for the file. In addition, scripts with the .mjs file extension can be used. They will be loaded with async/await symantics via the ES6 `import` statement.

### DIFF
--- a/bin/hubot.js
+++ b/bin/hubot.js
@@ -89,121 +89,121 @@ if (options.create) {
   console.error('See https://github.com/github/hubot/blob/master/docs/index.md for more details on getting started.')
   process.exit(1)
 }
+(async () => {
+  const robot = Hubot.loadBot(undefined, options.adapter, options.enableHttpd, options.name, options.alias)
+  await robot.loadAdapter(options.adapter)
+  if (options.version) {
+    console.log(robot.version)
+    process.exit(0)
+  }
 
-const robot = Hubot.loadBot(undefined, options.adapter, options.enableHttpd, options.name, options.alias)
+  if (options.configCheck) {
+    loadScripts()
+    console.log('OK')
+    process.exit(0)
+  }
 
-if (options.version) {
-  console.log(robot.version)
-  process.exit(0)
-}
+  robot.adapter.once('connected', () => loadScripts().then().catch(err => console.error(err)))
 
-if (options.configCheck) {
-  loadScripts()
-  console.log('OK')
-  process.exit(0)
-}
+  robot.run()
 
-robot.adapter.once('connected', loadScripts)
+  async function loadScripts () {
+    await robot.load(pathResolve('.', 'scripts'))
+    await robot.load(pathResolve('.', 'src', 'scripts'))
 
-robot.run()
+    loadHubotScripts()
+    loadExternalScripts()
 
-function loadScripts () {
-  robot.load(pathResolve('.', 'scripts'))
-  robot.load(pathResolve('.', 'src', 'scripts'))
-
-  loadHubotScripts()
-  loadExternalScripts()
-
-  options.scripts.forEach((scriptPath) => {
-    if (scriptPath[0] === '/') {
-      return robot.load(scriptPath)
+    for await (const scriptPath of options.scripts) {
+      if (scriptPath[0] === '/') {
+        return await robot.load(scriptPath)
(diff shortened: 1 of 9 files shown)


docs/adapters/development.md
```diff
@@ -94,5 +94,11 @@ There is a an open issue in the node community around [npm linked peer dependenc
   }
   ```
 3. Now try running `hubot -a sample` again and see that the imports are properly loaded.
-4. Once this is working properly, you can build out the functionality of your adapter as you see fit.  Take a look at some of the other adapters to get some ideas for your implementation.
+4. Once this is working properly, you can build out the functionality of your adapter as you see fit. Take a look at some of the other adapters to get some ideas for your implementation.
   - Once packaged and deployed via `npm`, you won't need the dependency in `hubot` anymore since the peer dependency should work as an official module.
+
+# Or, just load the adapter file from local disk
```
xurizaemon, line 100: Suggest to remove "just" here, often intended to represent a step as small, but covers for minimising complexity in documentation.

src/robot.js
```diff
@@ -355,21 +353,39 @@ class Robot {
...
   //
   // Returns nothing.
-  loadFile (filepath, filename) {
+  async loadFile (filepath, filename) {
     const ext = path.extname(filename)
-    const full = path.join(filepath, path.basename(filename, ext))
+    const full = path.join(filepath, path.basename(filename))
 
     // see https://github.com/hubotio/hubot/issues/1355
     if (['.js', '.mjs', '.coffee'].indexOf(ext) == -1) { // eslint-disable-line
```
xurizaemon, line 361: Should we consider .cjs here also?

Should this array be somewhere more global so it can be re-used, or does it belong here only?

xurizaemon (commented):
@joeyguerra I see there are three changes here -

1. Support for ES6 module (.mjs) adapters
2. async loading
3. Loading local path adapters

Do they belong in a single PR or should they be separated?

If the async/await changes can be separated from the others, I suggest we should do so. This would also make the change more visible. I see the adapter tests are changed here so I wonder if downstream adapters will need to update their test coverage to introduce the async change.

(Perhaps it would be good for Hubot to have test coverage which consumes a few popular adapters, which might catch such cases?)

If these belong in the same PR because they can't be proposed separately, should the commit history be rebased to land the changes separately?

bin/hubot.js
```diff
@@ -89,121 +89,121 @@ if (options.create) {
...
-      scriptsPath = pathResolve('node_modules', 'hubot-scripts', 'src', 'scripts')
-      robot.loadHubotScripts(scriptsPath, scripts)
-    } catch (error) {
-      const err = error
-      robot.logger.error(`Error parsing JSON data from hubot-scripts.json: ${err}`)
-      process.exit(1)
-    }
+      hubotScriptsWarning = 'Loading scripts from hubot-scripts.json is deprecated and ' + 'will be removed in 3.0 (https://github.com/github/hubot-scripts/issues/1113) ' + 'in favor of packages for each script.\n\n'
 
-    hubotScriptsWarning = 'Loading scripts from hubot-scripts.json is deprecated and ' + 'will be removed in 3.0 (https://github.com/github/hubot-scripts/issues/1113) ' + 'in favor of packages for each script.\n\n'
```
xurizaemon, line 149: Not for this PR, but noting since the line is changed here: this deprecated behaviour should become an error now we've passed 3.0?

joeyguerra:
I like the idea of splitting it up into multiple PRs. async might be required for loading '.mjs' files as adapters. But Let me see what I can do.

src/robot.js
technicalpickles: We previously used `require.extensions` which meant not needing to hard-code this. I'm not sure if anything replaced it.

joeyguerra:
It was too difficult for me to tease out the changes so I'm going to create separate PRs for each one and close this one, using it as reference. I'm already seeing some benefits in this approach. It forced me to hone an efficient development setup, come up with ideas on testing adapter integration, and add additional tests for adapter loading.

Issue #1624, opened 2023-05-11
Loading an adapter from local folder


joeyguerra:
I want to load an adapter that is in the same folder. Notice this line in `robot.js`. It doesn't allow me to specify a path to a local file to load as an adapter.

```javascript
      const path = Array.from(HUBOT_DEFAULT_ADAPTERS).indexOf(adapter) !== -1 ? `${this.adapterPath}/${adapter}` : `hubot-${adapter}`
```

I want to do something like `hubot -a ./src/adapters/MyCustomAdapter.mjs`.

PR #1635, opened 2023-06-03
feat(adapter): Load adapter from local disk

joeyguerra:
- load a .mjs file as an adapter (ES6 import
- load a .coffe, .js or .cjs file from local disk as an adapter (e.g. -f ./adapters/CustomAdapter.coffee, -f ./adapters/CustomAdapter.js, -f ./adapters/CustomAdapter.cjs)
- introduce new cli argument -f (file) that specifies the adapter path and file to load the adapter, e.g. -f ./adapters/CustomAdapter.mjs

BREAKING CHANGE: ES6 import is async, thus loadAdapter was changed to an async function.

#1624

### DIFF
--- a/bin/hubot.js
+++ b/bin/hubot.js
@@ -101,7 +101,6 @@ function loadScripts () {
   robot.load(pathResolve('.', 'scripts'))
   robot.load(pathResolve('.', 'src', 'scripts'))
 
-  loadHubotScripts()
   loadExternalScripts()
 
   options.scripts.forEach((scriptPath) => {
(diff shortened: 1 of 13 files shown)


joh6nn:
I can't say that it's good, only that it isn't obviously egregiously bad :shrug: If nobody else is able to step in for a real review in the next day or so, message me and we can do a quick walkthrough of it on like Slack or Discord or whatever works

joeyguerra:
Sounds good. Thank you.

docs/adapters/development.md
```diff
@@ -33,7 +33,7 @@ class Sample extends Adapter {
     }
     run() {
         this.robot.logger.info('Run')
-        this.emit('connected')
+        this.emit('connected') // this line is required. Scripts won't get loadded if this event never fires.
```
xurizaemon, line 36: s/loadded/loaded/

For clarity, let's rephrase this to phrase positively instead of with a double negative, eg "The 'connected' event is required to trigger loading of Hubot scripts."

bin/hubot.js
```diff
@@ -8,7 +8,8 @@ const OptParse = require('optparse')
 const Hubot = require('..')
 
 const switches = [
-  ['-a', '--adapter ADAPTER', 'The Adapter to use'],
+  ['-a', '--adapter ADAPTER', 'The Adapter to use, e.g. "shell" (to load the default hubot shell adapter)'],
+  ['-p', '--path PATH', 'Path to adapter directory, e.g. "./adapters" (Note: if this is set, the adapter option will be used as the file name to load from this folder)'],
```
xurizaemon, line 12: With this change, the `-a` behaves one way when `-p` is active, and another way when it doesn't. This feels like it exposes the user to complexity in order to save addition of an extra switch. I'd suggest to be explicit both internally and externally.

I recommend that each switch should have a single defined purpose, rather than switching behaviour based on context. (I'm trying to think of another CLI tool which demonstrates the behaviour here, and can't, which to me suggests this isn't the way to do it.)

I would be more comfortable with a separate switch for "full path to adapter", or to make `-p` accept a full path to the adapter and override `-a` (that feels less complex to me).

xurizaemon (changes requested):
Commit c7cceb2af0a37114f8e41a9b00e160b750dff7d4 introduces a breaking change which was overdue to be removed.

However the commit message does not use semantic-release syntax to signal this, and it should.

Please use `git rebase -i`  to amend the commit message.

(Quick review this morning 'cos I got interrupted - sorry!)

bin/hubot.js
joeyguerra: I concur. I'll take another stab at it.

bin/hubot.js
```diff
@@ -26,7 +27,7 @@ const options = {
   enableHttpd: process.env.HUBOT_HTTPD !== 'false',
   scripts: process.env.HUBOT_SCRIPTS || [],
   name: process.env.HUBOT_NAME || 'Hubot',
-  path: process.env.HUBOT_PATH || '.',
+  path: process.env.HUBOT_PATH || undefined,
```
technicalpickles, line 30: I think this is what would happen if it's not defined, so might not be necessary. But, it can be nice to be explicit.

bin/hubot.js
technicalpickles: I'd also add that `--path` is general enough that it wouldn't be super obvious

Perhaps we can re-use `--adapter` to accept a name of an adapter, or a directory? It'd be possible to check if the value given is a directory on the filesystem to distinguish the two.

technicalpickles (commented):
> BREAKING CHANGE: ES6 import is async, thus loadAdapter was changed to an async function.

Forgive me if I'm not up to date ES6, async, etc. What does this mean in practical terms? How does it affect existing users, scripts, and adapter packages?

joeyguerra:
@xurizaemon @technicalpickles I've incorporated ya'lls suggestions.
- Separated the concepts of loading an adapter from an artifact registry (i.e. npmjs via hubot-slack) from a file via -f ./adapters/CustomAdapter.{js|mjs|cs|coffee}.
- Replace double negative phrasing
- Updated commit message that removes loading scripts from external-scripts.json

PR #1645, opened 2023-06-11
docs: Fix link to example

WikiRik:
When updating from `master` to `main` this link was also changed, while that repo does not have a `main` branch


joeyguerra (approved):
Good catch. Thank you @WikiRik .

Issue #1658, opened 2023-08-08
Roadmap. The march to `v10`


joeyguerra:
# Updates to the Hubot Ecosystem

- [x] Modernize [`hubot-discord`](https://github.com/hubot-friends/hubot-discord)
- [x] Build a [`hubot-teams`](https://github.com/hubot-friends/hubot-ms-teams) adapter
- [-] ~Modernize [`hubot-google-hangouts-chat `](https://github.com/googleworkspace/hubot-google-hangouts-chat)~ Moving to a new issue so I can close this one.
- [-] ~Update [Mattermost Adapter](https://github.com/loafoe/hubot-matteruser)~ Moving to a new issue so I can close this one.
- [x] Modernize [`hubot-slack`](https://github.com/hubot-friends/hubot-slack)
- [x] Update `hubot-mock-adapter`
- [x] Update `hubot-redis-brain`
- [x] Update `hubot-help`
- [x] Update `huhot-diagnostics`
- [x] Update `hubot-rules`
- [x] Create a [`hubot-irc`](https://github.com/hubot-friends/hubot-irc) for the latest version of Hubot. On [NPM](https://www.npmjs.com/package/@hubot-friends/hubot-irc)
- [x] Hubot instance generator script
- [x] Write code to transform the markdown files into HTML to replace the https://hubot.github.com site because I don't have access to update it to the latest documentation
- [x] Create site with GitHub pages

# Hubot Proper

## `v10`
- [x] ~Remove coffee script support~ Going to take time with this one because there's a lot of coffee script scripts that I don't want to break.
- [x] Replace `optparse` with custom code
- [x] Replace `cline` with custom code

## `v9.x`
- [x] npx hubot --create myhubot instead of using yeoman

## `v9` #1666 
- [x] Make datastore async/await
- [x] Make robot.run async/await
- [x] Replace `mocha`, `chai` and `sinon` with Node [Test Runner](https://nodejs.org/api/test.html)
- [x] Bump Node version to 18.

## ~`v9`~ `v8.1.0` #1664 Not sure why this PR didn't result in a major version bump, but it didn't. 

- [x] Utilize async/await for Robot/Adapter APIs (i.e `reply`, `send`, etc...)
- [x] Replace `async` module with `async/await` (I think this will break lots of scripts)

## `v8` [PR](https://github.com/hubotio/hubot/pull/1657)

- [x] Enable using `.mjs` files as scripts

## `v7`

[...]

PR #1694, opened 2023-11-09
Remove Coffeescript

joeyguerra:
BREAKING CHANGES: Removed CoffeeScript support; .coffee files will no longer load
BREAKING CHANGES: Removed es2015.js file; import index.js file instead


AbdullahWins (commented):
Well Done! Keep up good work!

stahnma:
Why remove this and break lots of plugins in the ecosystem? I don't see this tied to an issue with discussion or anything like that.

joeyguerra:
A few thoughts:

Version 10.0.5 still supports coffeescript. For those still using hubot and coffeescripts, they should set the hubot peer dependency to version 10. 

I haven’t seen any  GitHub issues about breaking plugins. I’d love to know how many people are still using hubot. And hubot with coffeescripts.

Removing coffeescript was on the roadmap issue #1658 

Some background:
Development on hubot stalled sometime around 2017. I asked to be a maintainer and was added to the repo this year (at least I think it was this year). My objective was to modernize the code base, and try to pick up where the other maintainers left off with the goals in the evolution repo.

in my research, I found that almost every repo I found was last updated many years ago. Leaving me with the impression that the hubot ecosystem has gone stale. 

Being the 1 person maintaining it, I’ve been trying to reduce the maintenance surface area. Specifically reducing dependencies. I still get dependabot notifications for the yo generator repo. Gotta figure out what to do with that. 

I think there’s better adoption for ES6, which is my motivation to modernize hubot and evolve it to a new vision having been left stale for so long. 

With all those points in mind, I removed coffeescript in v11 as I try to incorporate some of the goals initially defined in the evolution repo and figure out where hubot now fits in todays universe.

Are you still using hubot? With coffeescripts? I’d love to hear your story.

PR #1718, opened 2024-04-02
Fix clobbered helper methods on msg object

PColuccio:
Helper methods on the `msg` object are being clobbered. This removes functions like reply, and random from `msg` passed to the callback.


joeyguerra:
Do you happen to have a test to highlight the scenario?

joeyguerra:
I wrote a failing test #1721 and fixed it with a different approach. Thank you for submitting this.

PR #1720, opened 2024-04-05
fix:  is not a function in CatchAll listener

### LINKED ISSUE #1721: response.reply is not a function in CatchAll listener
Someone submitted PR #1718. After no response about a test, I wrote a failing test to highlight the scenario.

joeyguerra:
fixes #1721

### DIFF
--- a/src/Robot.mjs
+++ b/src/Robot.mjs
@@ -229,7 +229,7 @@ class Robot {
     }
 
     this.listen(isCatchAllMessage, options, async msg => {
-      await callback(msg.message)
+      await callback(msg)
     })
   }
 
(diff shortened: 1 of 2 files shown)


PColuccio:
@joeyguerra I don't believe the `msg.message` object will be what you expect in this case though. In the original Hubot they were replacing the `msg.message` with `msg.message.message` although I haven't dug in to find the reason.

joeyguerra:
msg is a response. msg.message is an instance of CatchAllMessage. msg.message.message is an instance of TextMessage. Do you think you can try it? I assume you were getting errors in your code, and that led you to submitting a fix?

PR #1747, opened 2024-11-20
Minor change on documentation of Message and casing of the adapterName

KeesCBakker:


KeesCBakker:
Technically `done` and `room` are also part of every message.

joeyguerra:
Thank you for contributing.

We’re using semantic to create releases and publish to npm. Which is driven by certain syntax in he commit messages. If you revise your last commit message to be of the following format, a release will be triggered when this PR is smerged. 

fix: comment about why the change is made.

PR #1748, opened 2025-01-01
Update LICENSE.md, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2024 GitHub Inc.
+Copyright (c) 2011-2025 GitHub Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


xurizaemon:
There's little value to retaining the end date for a copyright range here. I suggest we instead replace this instead with `Copyright (c) 2011 GitHub Inc`. That would instead eliminate the periodic maintenance task of updating the copyright date for future years.

xurizaemon:
NB: [This user](https://github.com/JasonnnW3000) has 1.5K similar PRs across a number of repositories.

joeyguerra:
> NB: [This user](https://github.com/JasonnnW3000) has 1.5K similar PRs across a number of repositories.

I don't know whether to be appalled or impressed.

xurizaemon:
Mmm. I think the user can fill out their commit graph without us 😁 

@JasonnnW3000 project maintenance has associated costs - a PR like this can create work for maintainers, and automating PRs across thousands of projects is something you should think carefully about in future. I will assume your effort is well-motivated, but I'd caution you to consider whether it's useful ... or just generating work.

The value of updating the copyright year annually is next to zero IMO. I'd argue projects should remove the range end entirely, and we may do that in this project.

Perhaps consider whether "going deep, not broad" would be a more effective strategy to contribute, if that's your goal?

PR #1752, opened 2025-02-11
Delete LICENSE.md

Mopineda12:


xurizaemon:
No?

No indication here of why you're proposing to remove the license, but the change proposed is incomplete anyway as the file is referenced elsewhere in the project.

Please have a look at https://github.com/hubotio/hubot/blob/main/CONTRIBUTING.md 

@mopineda12 PRs should be clear about the "why" of a change, not just the change itself. Even if the thinking feels clear to you, it needs to be explicit for others.

I will close this incomplete PR. You're welcome to reopen once you have something which is more ready for review, of course.

I will note that changing the license of this project (if that's the intent?) would be an decision for the copyright holders, not the maintainers.

PR #1758, opened 2025-03-13
feat: load typescript scripts

tmm1:
when using modern nodejs with --experimental-strip-types, typescript can be loaded directly

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 20.x
+          - 23.x
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


joeyguerra:
Will you please add a unit test for this?

tmm1:
The test fails because it requires node v22 and the `--experimental-strip-types` flag. The patch is working on my hubot with those requirements.

joeyguerra:
Sounds like you'll have to change the pipeline to use Node 23.
https://github.com/tmm1/hubot/blob/51b60f0f1565971f4f665c0685f6b9ae3e0ce04f/.github/workflows/pipeline.yml#L45

and update the test command to pass the `--experimental-strip-types` flag. Try this -> `npm test -- --experimental-strip-types`

tmm1:
> Sounds like you'll have to change the pipeline to use Node 23.

Thanks ✅

PR #1773, opened 2025-06-12
Update license year

stagnation:
This is a linked example from https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository And last commit updated it to 2024 (last year).

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2024 GitHub Inc.
+Copyright (c) 2011-2025 GitHub Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


joeyguerra:
gonna wait on this to hear back from GitHub.

xurizaemon:
(Intentionally commenting and leaving this closed, thanks @joeyguerra)

Previously:

- https://github.com/hubotio/hubot/pull/1711
- https://github.com/hubotio/hubot/pull/1681
- https://github.com/hubotio/hubot/pull/1539
- https://github.com/hubotio/hubot/pull/860

and some longer thoughts at https://chris.bur.gs/copyright-end-date/

This PR is not the one I'm referring to in that post above. @stagnation, thanks for your effort.

I think the best solution would be to drop the end date, so we never have to do this again.

For this project, it's complicated by the license holder having disengaged from the project for several years. If we were the copyright owners, it would be easy to decide to remove the end date. Since we can't speak for Github, and Github haven't engaged publicly with this project in some time, it's less clear if doing that is OK. If we can't decide to change the copyright end date, my alternative proposal is that we:

1. Agree that until we decide otherwise, we won't update copyright end dates.
2. Add an HTML comment to the LICENSE.md which says, "Please do not PR changes to the license dates unless you represent the copyright owner. See CONTRIBUTING.md"
3. Note this in CONTRIBUTING.md also, directing people to how else they can contribute.

Just some thoughts, since we'll presumably get another PR like this in six months 😉

stagnation:
Thanks! That makes a lot of sense and certainly sounds like a thorny issue. Many well-wishing and subtly pedantic people sees the opportunity to make something better :). Feeling somewhat emboldened by the ease of edit->PR in github now.
Hope I did not cause any frustration.

Great explanation in the blog post as well :+1: 
Have a nice day. //Nils

PR #1811, opened 2026-01-11
Update Message_test.mjs

sudipta-dotcom:
I added a new key or property in the object user as displayName...so that I can further use it while adressing the user by their display name instead of name and for  further changes in the file index.js.


sudipta-dotcom:
Sir,could you please help me to select some new issues beginner
friendly...I am contributing and want to contribute further...related to
javascript.

